PR #54, opened 2023-06-16
WB-380 | Manual payment

diogo-wopta:
Update logic for allowing manual payment of single transaction:
- must not be paid;
- must be the next installment for the related policy;
- the policy must be signed

Update Policy if was not paid as in callback/payment:
- move contract and attach it to the policy
- create/update user
- update policy fields
- send email with the contract to the user

We've also update the global handling of errors on the router. Now instead of returning a status code 500 (that with the interception of the appcheck function arrives as a 200 at the Frontend), we return a 200 with an error message

### DIFF
--- a/lib/router.go
+++ b/lib/router.go
@@ -61,11 +61,11 @@ func (router RouteData) Router(w http.ResponseWriter, r *http.Request) {
 	}
 	if e != nil {
 		log.Println("Router error")
+		log.Println(e.Error())
 		// Temporary work-around while test-function is not rewritten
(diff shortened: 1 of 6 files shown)


lib/router.go
```diff
@@ -61,9 +61,12 @@ func (router RouteData) Router(w http.ResponseWriter, r *http.Request) {
 	}
 	if e != nil {
 		log.Println("Router error")
-		log.Println(e.Error())
```
Yousef-Hammar, line 64: should we keep this log?

Yousef-Hammar (approved):

PR #56, opened 2023-06-19
Refactor to user invite creation

nahidalgo:
Added fiscalCode, name and surname to invite creation.
Extra controls to check if the user already exists.

### DIFF
--- a/user/consume-invite.go
+++ b/user/consume-invite.go
@@ -65,7 +65,10 @@ func ConsumeInvite(inviteUid, password, origin string) (bool, error) {
 		Surname:    invite.Surname,
 	}
 
-	lib.SetFirestore(usersCollectionName, user.Uid, user)
+	err = lib.SetFirestoreErr(usersCollectionName, user.Uid, user)
+	if err != nil {
+		return false, err
+	}
 
 	// update the user custom claim
 	lib.SetCustomClaimForUser(user.AuthId, map[string]interface{}{
--- a/user/consume-invite.go
+++ b/user/consume-invite.go
@@ -43,7 +43,7 @@ func ConsumeInvite(inviteUid, password, origin string) (bool, error) {
 
 	// Check if invite is not consumed nor expired
 	if invite.Consumed || time.Now().UTC().After(invite.Expiration) {
-		return false, errors.New("Invite consumed or expired")
+		return false, errors.New("invite consumed or expired")
 	}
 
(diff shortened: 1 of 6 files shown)


user/create-invite.go
```diff
@@ -24,32 +27,58 @@ func CreateInviteFx(w http.ResponseWriter, r *http.Request) (string, interface{}
 		return `{"success": false}`, `{"success": false}`, nil
 	}
 
-	if inviteUid, ok := CreateInvite(createInviteRequest.Email, createInviteRequest.Role, r.Header.Get("Origin"), creatorUid); ok {
+	if inviteUid, err := CreateInvite(createInviteRequest.Name, createInviteRequest.Surname, createInviteRequest.FiscalCode, createInviteRequest.Email, createInviteRequest.Role, r.Header.Get("Origin"), creatorUid); err != nil {
```
Yousef-Hammar: Pass entire `CreateInviteRequest` struct as input to `CreateInvite` function

user/consume-invite.go
```diff
@@ -56,12 +56,15 @@ func ConsumeInvite(inviteUid, password, origin string) (bool, error) {
...
+		Uid:        userRecord.UID,
+		AuthId:     userRecord.UID,
+		Role:       invite.Role,
+		FiscalCode: invite.FiscalCode,
+		Name:       invite.Name,
+		Surname:    invite.Surname,
 	}
-	
+
 	lib.SetFirestore(usersCollectionName, user.Uid, user)
```
diogo-wopta: Shouldn't we use the new function SetFirestoreErr and handle eventual errors?

user/user.go
```diff
@@ -169,3 +169,14 @@ func SetUserIntoPolicyContractor(policy *models.Policy, origin string) {
 		lib.CheckError(err)
 	}
 }
+
+func GetAuthUserByMail(mail string) (models.User, error) {
+	var user models.User
+
+	authId, err := lib.GetAuthUserIdByEmail(mail)
```
diogo-wopta, line 176: Even though both have the same value (as seen on consume-invite.go lines 60, 61), the function GetAuthUserIdByEmail returns the UID, not the authId.

nahidalgo: The user in `GetAuthUserIdByEmail` is not a `User` from models, instead, it is a Firebase User, which the UID is the same as our `models.User.Uid` and `models.User.AuthId`.

diogo-wopta (approved):

Yousef-Hammar (approved):

PR #58, opened 2023-06-22
Wb 376

pierlucatonioloreply:
Create Query Users API

Follow the idea of the Get Policies with query, in the broker domain, file get-policies.go

### DIFF
--- a/user/get-users.go
+++ b/user/get-users.go
@@ -1,0 +1,80 @@
+package user
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"time"
+
+	lib "github.com/wopta/goworkspace/lib"
+	models "github.com/wopta/goworkspace/models"
+)
+
+func GetUsersFx(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
+	var (
+		req        GetUsersReq
+		response   GetUsersResp
+		limitValue = 10
+	)
+	log.Println("GetUsers")
+
+	body := lib.ErrorByte(io.ReadAll(r.Body))
+	err := json.Unmarshal(body, &req)
+	lib.CheckError(err)
+
+	if len(req.Queries) == 0 {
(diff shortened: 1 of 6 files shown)


user/get-users.go
```diff
@@ -0,0 +1,79 @@
...
+type GetUsersReq struct {
+	Queries []struct {
+		Field string      `json:"field"`
+		Op    string      `json:"op"`
+		Value interface{} `json:"value"`
+		Type  string      `json:"type"`
+	} `json:"queries,omitempty"`
+	Limit int `json:"limit"`
+	Page  int `json:"page"`
+}
```
nahidalgo: This is equal to the `GetPoliciesReq`. I think there should be a Query struct in models. What do you think @Yousef-Hammar  @diogo-wopta

nahidalgo: Add these structs on the top of the file

user/user.go
```diff
@@ -81,6 +81,12 @@ func User(w http.ResponseWriter, r *http.Request) {
 				Method:  http.MethodPatch,
 				Roles:   []string{models.UserRoleAdmin},
 			},
+			{
+				Route:   "/user/v1",
```
nahidalgo: The domain already has user, so you can leave only `v1`
```suggestion
				Route:   "/v1",
```

nahidalgo (changes requested):

user/user.go
```diff
@@ -81,6 +81,12 @@ func User(w http.ResponseWriter, r *http.Request) {
...
+				Handler: GetUsersFx,
+				Method:  http.MethodPost,
+				Roles:   []string{models.UserRoleAll},
```
nahidalgo: Here, the role should be `models.UserRoleAdmin`

nahidalgo (dismissed):

Yousef-Hammar (approved):

diogo-wopta (approved):

diogo-wopta (approved):

PR #59, opened 2023-06-23
Added audit-log.go

ruixuan-xu:


models/audit-log.go
```diff
@@ -0,0 +1,90 @@
...
+// The AuditLog is a data structure that stores request's data.
+// In particular, the Payload should contain data from the request's body.
+//
+// Example:
+//
+//	a := ParseHttpRequest(r)
+//	a.SaveToBigQuery()
+type AuditLog struct {
+	Payload  string    `bigquery:"payload"`
+	Date     time.Time `bigquery:"date"`
```
diogo-wopta: In BigQuery Date is saved as `civil.DateTime`

models/audit-log.go
```diff
@@ -0,0 +1,90 @@
...
+	UserUid  string    `bigquery:"userUid"`
+	Method   string    `bigquery:"method"`
+	Endpoint string    `bigquery:"endpoint"`
+	Role     string    `bigquery:"role"`
+}
+
+func ParseHttpRequest(r *http.Request) (AuditLog, error) {
+	if r.Body != nil {
+		return AuditLog{}, fmt.Errorf("no payload found: body is empty!")
+	}
```
diogo-wopta: The control is inverted. Regardless it could be in the future that we'd like to audit activities that does not require a payload. In those cases we should fill the field as an empty string

models/audit-log.go
```diff
@@ -0,0 +1,90 @@
...
+
+	defer r.Body.Close()
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		return AuditLog{}, fmt.Errorf("cannot retrieve the payload: %v", err)
+	}
+
+	userUid := r.Header.Get("userUid")
```
diogo-wopta: The userUid wont't be available from the Header. We have to extract it from the authorization token. Check `lib.VerifyUserIdToken`

models/audit-log.go
```diff
@@ -0,0 +1,90 @@
...
+	token, err := lib.VerifyUserIdToken(idToken)
+	if err != nil {
+		return "", fmt.Errorf("verify ID token error: %s", err)
+	}
+	if token.Claims["role"] == nil {
+		return "", fmt.Errorf("user role not set")
+	}
+	role := token.Claims["role"].(string)
+	return role, nil
+}
```
diogo-wopta: When we merge the WB-367 branch there will be a function `GetAuthTokenFromIdToken` that might come handy for getting a users Uid and Role from its token.

models/audit-log.go
```diff
@@ -0,0 +1,90 @@
...
+
+	userUid := r.Header.Get("userUid")
+	if userUid == "" {
+		return AuditLog{}, fmt.Errorf("cannot retrieve the user's UID. Header (userUid) is empty")
+	}
+
+	role, err := getRole(r)
+	if err != nil {
+		return AuditLog{}, fmt.Errorf("cannot retrieve the role: %s", err)
+	}
```
diogo-wopta: There is a function `lib.GetUserRoleFromIdToken`

diogo-wopta (approved):

PR #63, opened 2023-07-03
WB-437 | Broker emit approval

diogo-wopta:
Refactor of the emit function, splitting the different actions (base, sign and pay) and adding the new approval action.

We need to refine what would be the correct status updates if we start making those actions non-linear, ex.: payment only, signing only, not sending email to user etc.


broker/emit.go
```diff
@@ -9,98 +9,149 @@ import (
...
+}
+
+func Emit(policy *models.Policy, request EmitRequest, origin string) EmitResponse {
+	var responseEmit EmitResponse
+
+	firePolicy := lib.GetDatasetByEnv(origin, "policy")
+	guaranteFire := lib.GetDatasetByEnv(origin, "guarante")
+	policy.Uid = request.Uid // we should enforce the setting of the ID on proposal
+
+	if policy.IsReserved {
```
Yousef-Hammar, line 72: change this to `if policy.IsReserved && policy.Status == "WaitForApproval"`

PR #67, opened 2023-07-05
Implemented a sellable vehicle

ruixuan-xu:


sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+			return fmt.Errorf("The coverage has duration 0")
+		}
+		if pd != coverage {
+			return fmt.Errorf("wrong policy duration! it should be %d, we've got %d", coverage, pd)
+		}
+	}
+	return nil
+}
+
+func sliceContains[T comparable](s []T, o T) bool {
```
diogo-wopta, line 145: This function is already declared in lib.SliceContains

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+	for _, e := range s {
+		if o == e {
+			return true
+		}
+	}
+	return false
+}
```
diogo-wopta: This function already exists in lib.SliceContains

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+// Function for exposing the sellability function.
+// It takes a request in which the body contains the policy in JSON format,
+// and then it returns the appropriate product.
+func VehicleHandler(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
+	var (
+		policy models.Policy
+		err    error
+	)
+
+	log.Println("Vehicle")
```
diogo-wopta: use same name as function
log.Println("[GapFx] Handler start")

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/wopta/goworkspace/lib"
+	"github.com/wopta/goworkspace/models"
+	prd "github.com/wopta/goworkspace/product"
+)
+
+// Function for exposing the sellability function.
+// It takes a request in which the body contains the policy in JSON format,
+// and then it returns the appropriate product.
+func VehicleHandler(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
```
diogo-wopta: rename in GapFx

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+	product, productJson, err := Vehicle(authToken.Role, policy)
+
+	return productJson, product, err
+}
+
+// Given a policy that shoudl contain the Vehicle asset, then it returns:
+//   - the product or parts of it depending on the sellable rules
+//   - its representation in json
+//   - and an eventual error
+func Vehicle(role string, p models.Policy) (models.Product, string, error) {
```
diogo-wopta: rename in Gap

diogo-wopta: the function should return a product and an error. The marshaling of the product and its output as a string should be done on the handler function.
This will be the function used by the quote/gap endpoint as to not make an unnecessary http call between functions.

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+	return nil
+}
+
+func validatePolicy(p models.Policy) error {
+	if len(p.Assets) == 0 {
+		return fmt.Errorf("no assets found")
+	}
+	if p.Assets[0].Vehicle == nil {
+		return fmt.Errorf("no vehicle found")
+	}
```
diogo-wopta, line 136: add check for p.Assets[0].Person

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+
+	v := p.Assets[0].Vehicle
+	pd := elapsedYears(p.StartDate, p.EndDate)
```
diogo-wopta: there is a "similar" function in the quote PR. If the logic is/should be the same we should extract it into lib

sellable/vehicle.go
```diff
@@ -0,0 +1,177 @@
...
+		coverage := 5 - decrease
+
+		if coverage <= 0 {
+			return fmt.Errorf("The coverage has duration 0")
+		}
+		if pd != coverage {
+			return fmt.Errorf("wrong policy duration! it should be %d, we've got %d", coverage, pd)
+		}
+	}
+	return nil
```
diogo-wopta: use constants instead of values.
Ex.:
maxAgeAtStartDate = 5
maxAgeAtEndDate = 8
minimumVehiclePriceValue = 4000
maximumVehiclePriceValue = 120000

sellable/vehicle.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	v := p.Assets[0].Vehicle
+	pd := lib.ElapsedYears(p.StartDate, p.EndDate)
+
+	maxAgeFullCoverageBD := v.RegistrationDate.AddDate(maxAgeFullCoverage, 0, 0)
+	if time.Now().Before(maxAgeFullCoverageBD) {
+		if pd != maxCoverage {
+			return fmt.Errorf("wrong policy duration! it should be 5, we've got %d", pd)
+		}
```
Yousef-Hammar, line 145: this should be > and not != because you can have a duration up to 5 years if the car has less than 3 years old

sellable/vehicle.go
```diff
@@ -0,0 +1,158 @@
...
+	} else {
+		decrease := lib.ElapsedYears(maxAgeFullCoverageBD, time.Now())
+		coverage := maxCoverage - decrease
+
+		if coverage <= 0 {
+			return fmt.Errorf("The coverage has duration 0")
+		}
+		if pd != coverage {
+			return fmt.Errorf("wrong policy duration! it should be %d, we've got %d", coverage, pd)
+		}
```
Yousef-Hammar, line 155: this should be > (same reason as above)

sellable/vehicle.go
```diff
@@ -0,0 +1,158 @@
...
+		return "{}", nil, fmt.Errorf("cannot read body: %v", err)
+	}
+	if err = json.Unmarshal(bytes, &policy); err != nil {
+		return "{}", nil, err
+	}
+
+	authToken, err := models.GetAuthTokenFromIdToken(r.Header.Get("Authorization"))
+	lib.CheckError(err)
+	product, err := Vehicle(authToken.Role, policy)
+
```
Yousef-Hammar, line 49: check error here

sellable/vehicle.go
```diff
@@ -0,0 +1,158 @@
...
+	if v.MainUse != "private" {
+		return fmt.Errorf("the vehicle is not private")
+	}
+
+	car_types := []string{"auto", "autocarro", "suv"}
+	if !lib.SliceContains(car_types, v.VehicleType) {
+		return fmt.Errorf("The vehicle type is not in: %v", car_types)
+	}
+
+	anniversary := v.RegistrationDate.AddDate(maxAgeAtStartDate, 0, 0)
```
Yousef-Hammar, line 107: put this in unique if statement

ruixuan-xu: I don't understand, every singular call is checked in its own "unique" if statement

PR #78, opened 2023-07-18
WB 477/478 | Policy annulment and refund api

diogo-wopta:
TODO: agree with frontend on what should be the payload. As it is described in the wireframes it's misleading: one cannot create an annulment request with a mismatched code/description/refund. Each code is unique on its actions/side-effects

TODO: add save to BigQuery for modified policy and transactions

TODO: review fabrick expire integration

TODO: rename refundTransactions into createRefundTransactions


diogo-wopta:
Too outdated and done with different requirements

PR #84, opened 2023-07-24
[HOTFIX] Changed how we calculate the Wise policy price to have also the payment split

nahidalgo:
- Fixed problem in models/wise-guarantee.go, where the GuaranteeValue was being cleared because it was declared inside the for loop, and not outside as it should be
- Added semestral payment split because in wise we have some semestral policies and this will be visualized in app
- [TEMP] Added an offset to the policy EndDate (only for wise policies) so we can visualize the date correctly on the app. This requires a deeper understanding for all the dates, because we are saving them incorrectly.

### DIFF
--- a/models/wise-complete-policy.go
+++ b/models/wise-complete-policy.go
@@ -87,14 +87,14 @@ func updatePolicyPrice(policy *Policy, wisePolicy *WiseCompletePolicy) {
 	if wisePolicy.Contract.InstalmentTypeCode == WiseMonthlyPaymentSplitCode {
 		policy.PaymentSplit = string(PaySplitMonthly)
 		policy.PriceGross = policy.PriceGrossMonthly * 12
-		policy.PriceNett = policy.PriceNett * 12
+		policy.PriceNett = policy.PriceNettMonthly * 12
 		policy.TaxAmount = policy.TaxAmountMonthly * 12
 		return
 	}
 	if wisePolicy.Contract.InstalmentTypeCode == WiseSemestralPaymentSplitCode {
 		policy.PaymentSplit = string(PaySplitSemestral)
 		policy.PriceGross = policy.PriceGrossMonthly * 2
-		policy.PriceNett = policy.PriceNett * 2
+		policy.PriceNett = policy.PriceNettMonthly * 2
 		policy.TaxAmount = policy.TaxAmountMonthly * 2
 		return
 	}
--- a/models/wise-complete-policy.go
+++ b/models/wise-complete-policy.go
@@ -32,7 +32,7 @@ func (wisePolicy *WiseCompletePolicy) ToDomain() Policy {
 	policy.Uid = fmt.Sprintf("wise:%d", wisePolicy.Id)
 	policy.Contractor = *wisePolicy.Contractors[0].Registry.ToDomain()
 	policy.CodeCompany = wisePolicy.PolicyNumber
-	
+
 	updatePolicyPrice(&policy, wisePolicy)
 
 	policy.ProductVersion = "v1"
--- a/models/wise-complete-policy.go
+++ b/models/wise-complete-policy.go
@@ -3,6 +3,8 @@ package models
 import (
 	"fmt"
 	"time"
+
+	"github.com/wopta/goworkspace/lib"
 )
 
 type WiseCompletePolicy struct {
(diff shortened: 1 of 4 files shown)


models/wise-complete-policy.go
```diff
@@ -59,6 +62,44 @@ func (wisePolicy *WiseCompletePolicy) ToDomain() Policy {
...
+	}
+
+	policy.PriceGrossMonthly = wisePolicy.Contract.GrossAmount
+	policy.PriceNettMonthly = wisePolicy.Contract.NetAmount
+	policy.TaxAmountMonthly = policy.PriceGrossMonthly - policy.PriceNettMonthly
+
+	if wisePolicy.Contract.InstalmentTypeCode == WiseMonthlyPaymentSplitCode {
+		policy.PaymentSplit = string(PaySplitMonthly)
+		policy.PriceGross = policy.PriceGrossMonthly * 12
+		policy.PriceNett = policy.PriceNett * 12
```
nahidalgo: This should be 
```suggestion
		policy.PriceNett = policy.PriceNettMonthly * 12
```

models/wise-complete-policy.go
```diff
@@ -59,6 +62,44 @@ func (wisePolicy *WiseCompletePolicy) ToDomain() Policy {
...
+		policy.TaxAmount = policy.TaxAmountMonthly * 12
+		return
+	}
+	if wisePolicy.Contract.InstalmentTypeCode == WiseSemestralPaymentSplitCode {
+		policy.PaymentSplit = string(PaySplitSemestral)
+		policy.PriceGross = policy.PriceGrossMonthly * 2
+		policy.PriceNett = policy.PriceNett * 2
```
nahidalgo: Also here
```suggestion
		policy.PriceNett = policy.PriceNettMonthly * 2
```

models/wise-complete-policy.go
```diff
@@ -59,6 +62,44 @@ func (wisePolicy *WiseCompletePolicy) ToDomain() Policy {
 	return policy
 }
 
+func updatePolicyEndDate(policy *Policy, wisePolicy *WiseCompletePolicy) {
+	location, err := time.LoadLocation("Europe/Rome")
+	lib.CheckError(err)
+	policyEndDate := wisePolicy.Contract.PolicyExpirationDate.In(location)
+	_, offset := policyEndDate.Zone()
+	policy.EndDate = policyEndDate.Add(time.Duration(time.Second * time.Duration(offset)))
+}
```
diogo-wopta, line 71: This probably relates to what we discussed in call earlier, but won't this code add the time offset two times?
[goplayground](https://go.dev/play/p/oYb2op9wtXx?v=goprev)

models/wise-complete-policy.go
```diff
@@ -59,6 +62,44 @@ func (wisePolicy *WiseCompletePolicy) ToDomain() Policy {
...
+		policy.PriceGross = policy.PriceGrossMonthly * 12
+		policy.PriceNett = policy.PriceNett * 12
+		policy.TaxAmount = policy.TaxAmountMonthly * 12
+		return
+	}
+	if wisePolicy.Contract.InstalmentTypeCode == WiseSemestralPaymentSplitCode {
+		policy.PaymentSplit = string(PaySplitSemestral)
+		policy.PriceGross = policy.PriceGrossMonthly * 2
+		policy.PriceNett = policy.PriceNett * 2
+		policy.TaxAmount = policy.TaxAmountMonthly * 2
```
diogo-wopta: Is this logic accurate?
I'd expect monthly * 6 or yearly / 2

nahidalgo: Yes, because we are saving the `*Semestral` price also in the `*Monthly` prices. Me and @Yousef-Hammar have discussed that it makes more sense to use the `PriceGross` as the annual value and `PriceGrossMonthly` as the related `PaymentSplit` value, be it `monthly`, `semestral` or `singleInstallment`. The thing that we need to do as a next step would be renaming the `*Monthly` prices to something more fitting. Do you agree @diogo-wopta?

nahidalgo: And also I don't think it makes sense to have a field for each `PaymentSplit` like, `PriceGrossSingleInstallment`, `PriceGrossSemestral`.

models/wise-complete-policy.go
nahidalgo: It adds only once, but it also changes the Timezone, in the example you sent:
- The time was `2009-11-10 23:00:00 +0000 UTC`
- When adding the offset the time became `2009-11-11 00:00:00 +0000 UTC` or, in the Rome timezone `2009-11-11 01:00:00 +0100 CET`

Yousef-Hammar (approved):

models/wise-complete-policy.go
diogo-wopta: Ok. A few options are:
PriceGrossSplit / PriceGrossRate / PriceGrossInstallment

Yousef-Hammar (approved):

PR #113, opened 2023-09-29
TG-191 | Network transactions

diogo-wopta:
TODO:
add audit for accounting POST/PUT operations


accounting/accounting.go
```diff
@@ -1,119 +1,45 @@
...
+	route := lib.RouteData{
+		Routes: []lib.Route{
+			{
+				Route:   "/network/transactions/v1/transaction/:transactionUid",
+				Handler: GetNetworkTransactionsFx,
+				Method:  http.MethodGet,
+				Roles:   []string{models.UserRoleAdmin, models.UserRoleManager},
+			},
+			{
+				Route:   "/network/transactions/v1/:uid",
```
Yousef-Hammar, line 30: should be changed to `/network/transactions/v1/transaction/:transactionUid` in order to set a standard with other endpoints in domain

PR #114, opened 2023-10-02
TG-194 | Network node Part 2

diogo-wopta:
- created network node model;
- updated policy with network node data;
- created SaveBigQuery method for NetworkNode;
- created GET mga/network/node/:uid;
- created POST mga/network/node;
- update simplo agent/agency data;
- ifon broker pass the network node to avoid retrieving it on all steps;

TODO:
- update create invite (v2);
- update consume invite (v2);
- update onboard user (v2);
- update companydata emitTrack;
- check all point where we check for agent/agency;
- create CRUD for network node in MGA;

NICE TO HAVE:
- create template for invite mail;

### DIFF
--- a/broker/get-transactions.go
+++ b/broker/get-transactions.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 
 	"github.com/wopta/goworkspace/lib"
 	"github.com/wopta/goworkspace/models"
@@ -36,17 +37,10 @@ func GetPolicyTransactionsFx(w http.ResponseWriter, r *http.Request) (string, in
 
 	userUid := authToken.UserID
 
-	switch authToken.Role {
-	case models.UserRoleAgent:
-		if policy.AgentUid != userUid {
-			log.Printf("[GetPolicyTransactionsFx] policy %s is not included in agent %s", policyUid, userUid)
-			return "", response, fmt.Errorf("agent %s unauthorized for policy %s", userUid, policyUid)
-		}
-	case models.UserRoleAgency:
-		if policy.AgencyUid != userUid {
-			log.Printf("[GetPolicyTransactionsFx] policy %s is not included in agency %s", policyUid, userUid)
-			return "", response, fmt.Errorf("agency %s unauthorized for policy %s", userUid, policyUid)
-		}
+	isAgentOrAgency := strings.EqualFold(authToken.Role, models.UserRoleAgent) || strings.EqualFold(authToken.Role, models.UserRoleAgency)
+	if isAgentOrAgency && policy.ProducerUid != userUid {
+		log.Printf("[GetPolicyTransactionsFx] policy %s is not included in %s %s portfolio", policyUid, authToken.Role, userUid)
+		return "", response, fmt.Errorf("%s %s unauthorized for policy %s", authToken.Role, userUid, policyUid)
 	}
 
 	transactions := transaction.GetPolicyTransactions(r.Header.Get("origin"), policyUid)
--- a/callback/bpmn.go
+++ b/callback/bpmn.go
@@ -16,13 +16,9 @@ import (
 )
 
 var (
-	origin        string
-	trSchedule    string
-	paymentMethod string
-	ccAddress     mail.Address
-	toAddress     mail.Address
-	fromAddress   mail.Address
-	networkNode   *models.NetworkNode
+	origin, trSchedule, paymentMethod string
+	ccAddress, toAddress, fromAddress mail.Address
+	networkNode                       *models.NetworkNode
 )
 
 const (
@@ -41,7 +37,7 @@ func runCallbackBpmn(policy *models.Policy, flowKey string) *bpmn.State {
 	)
 
 	fromAddress = mail.AddressAnna
-	channel := models.GetChannel(policy)
+	channel := policy.Channel
 	settingFile := fmt.Sprintf(settingFormat, channel)
 
 	log.Printf("[runCallbackBpmn] loading file for channel %s", channel)
@@ -65,6 +61,8 @@ func runCallbackBpmn(policy *models.Policy, flowKey string) *bpmn.State {
 		case models.AgencyChannel:
 			toAddress = mail.GetContractorEmail(policy)
 			ccAddress = mail.GetNetworkNodeEmail(networkNode)
+		case models.MgaChannel:
+			toAddress = mail.GetContractorEmail(policy)
 		default:
 			toAddress = mail.GetNetworkNodeEmail(networkNode)
 			ccAddress = mail.Address{}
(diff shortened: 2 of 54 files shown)


broker/get-transactions.go
```diff
@@ -36,17 +37,10 @@ func GetPolicyTransactionsFx(w http.ResponseWriter, r *http.Request) (string, in
...
-		}
-	case models.UserRoleAgency:
-		if policy.AgencyUid != userUid {
-			log.Printf("[GetPolicyTransactionsFx] policy %s is not included in agency %s", policyUid, userUid)
-			return "", response, fmt.Errorf("agency %s unauthorized for policy %s", userUid, policyUid)
-		}
+	isAgentOrAgency := strings.EqualFold(authToken.Role, models.UserRoleAgent) || strings.EqualFold(authToken.Role, models.UserRoleAgency)
+	if isAgentOrAgency && policy.ProducerUid != userUid {
+		log.Printf("[GetPolicyTransactionsFx] policy %s is not included in %s %s portfolio", policyUid, authToken.Role, userUid)
+		return "", response, fmt.Errorf("%s %s unauthorized for policy %s", authToken.Role, userUid, policyUid)
```
diogo-wopta, line 43: here we have to think this through. In theory, any upper node should be able to access the policy emitted from a descendent.

we might check policy.NetworkUid with loggedUser.NetworkUid and policy.ProducerUid with the traverse of the network to know if the node is in the accepted level

callback/bpmn.go
```diff
@@ -215,17 +219,9 @@ func updatePolicy(state *bpmn.State) error {
...
-		return err
-	}
-
-	// Update agent if present
-	err = models.UpdateAgentPortfolio(policy, origin)
-	if err != nil && err.Error() != "agent not set" {
-		log.Printf("[updatePolicy] ERROR UpdateAgentPortfolio %s", err.Error())
+	err = network.UpdateNetworkNodePortfolio(origin, policy, networkNode)
+	if err != nil {
+		log.Printf("[updatePolicy] error updating %s portfolio %s", networkNode.Type, err.Error())
```
diogo-wopta, line 224: how does it behave with e-commerce where we do not have a networkNode. Previously we would check for "agency/agent not set" error. We might need to call this update function only if we do have a node, otherwise change it to have an early non-error escape for a missing node.

document/contract.go
```diff
@@ -20,14 +20,14 @@ func ContractFx(w http.ResponseWriter, r *http.Request) (string, interface{}, er
 	defer r.Body.Close()
 	err := json.Unmarshal([]byte(req), &data)
 	lib.CheckError(err)
-	respObj := <-ContractObj(origin, data)
+	respObj := <-ContractObj(origin, data, nil)
```
diogo-wopta, line 23: ok if this function is not triggered externally, otherwise we should extract the node from the authToken

document/fpdf-data.go
```diff
@@ -72,37 +75,38 @@ func loadLifeBeneficiariesInfo(policy *models.Policy) ([]map[string]string, stri
...
+	case models.AgentNetworkNodeType:
+		policyProducer["name"] = strings.ToUpper(networkNode.Agent.Surname) + " " + strings.ToUpper(networkNode.Agent.Name)
+		policyProducer["ruiSection"] = networkNode.Agent.RuiSection
+		policyProducer["ruiCode"] = networkNode.Agent.RuiCode
+		policyProducer["ruiRegistration"] = networkNode.Agent.RuiRegistration.Format("02.01.2006")
+	case models.AgencyNetworkNodeType:
+		policyProducer["name"] = strings.ToUpper(networkNode.Agency.Name)
+		policyProducer["ruiSection"] = networkNode.Agency.RuiSection
+		policyProducer["ruiCode"] = networkNode.Agency.RuiCode
+		policyProducer["ruiRegistration"] = networkNode.Agency.RuiRegistration.Format("02.01.2006")
```
diogo-wopta, line 104: with the new structure we should retrieve the data from node.Agency.Manager

mga/create-network-node.go
```diff
@@ -0,0 +1,84 @@
...
+	Role        string                  `json:"role"`
+	NetworkCode string                  `json:"networkCode"`
+	ManagerUid  string                  `json:"managerUid,omitempty"`
+	ParentUid   string                  `json:"parentUid,omitempty"`
+	Agent       *models.AgentNode       `json:"agent,omitempty"`
+	Agency      *models.AgencyNode      `json:"agency,omitempty"`
+	Broker      *models.AgencyNode      `json:"broker,omitempty"`
+	Partnership *models.PartnershipNode `json:"partnership,omitempty"`
+	Products    []models.Product        `json:"products,omitempty"`
+}
```
diogo-wopta: if all the fields are in common with NetworkNode, shouldn't we accept one as request, considering that we map what we want in the createNetworkNode function?

mga/create-network-node.go
```diff
@@ -0,0 +1,84 @@
...
+		Agent:        request.Agent,
+		Agency:       request.Agency,
+		Broker:       request.Broker,
+		Partnership:  request.Partnership,
+		Products:     request.Products,
+		IsActive:     true,
+		CreationDate: now,
+		UpdatedDate:  now,
+	}
+}
```
diogo-wopta: use network.CreateNode(request)

models/network-node.go
```diff
@@ -39,56 +45,114 @@ type PartnershipNode struct {
...
-	return nil
+	log.Println("[NetworkNode.SaveBigQuery]")
+
+	nnJson, _ := json.Marshal(nn)
+
+	nn.Data = string(nnJson)
+	nn.BigCreationDate = lib.GetBigQueryNullDateTime(nn.CreationDate)
+	nn.BigUpdatedDate = lib.GetBigQueryNullDateTime(nn.UpdatedDate)
+	nn.parseBigQueryAgent()
+	nn.parseBigQueryAgency()
```
diogo-wopta: missing areaManager and broker functions

models/network-node.go
```diff
@@ -39,56 +45,114 @@ type PartnershipNode struct {
...
+		return
+	}
+
+	if nn.Agency.Address != nil {
+		nn.Agency.Address.BigLocation = lib.GetBigQueryNullGeography(
+			nn.Agency.Address.Location.Lng,
+			nn.Agency.Address.Location.Lat,
+		)
+	}
+	nn.Agency.BigRuiRegistration = lib.GetBigQueryNullDateTime(nn.Agency.RuiRegistration)
```
diogo-wopta: check new structure for possible missing fields ex.: manager

broker/get-transactions.go
diogo-wopta: We'll do it later

Yousef-Hammar (approved):

Yousef-Hammar (approved):

PR #121, opened 2023-10-16
Fix for lead being overwritten

nahidalgo:
https://tree.taiga.io/project/nahidalgo-wopta/issue/246

### DIFF
--- a/broker/lead.go
+++ b/broker/lead.go
@@ -19,6 +19,8 @@ func LeadFx(w http.ResponseWriter, r *http.Request) (string, interface{}, error)
 		policy models.Policy
 	)
 
+	w.Header().Add("content-type", "application/json")
+
 	log.Println("[LeadFx] Handler start -----------------------------------------")
 
 	origin = r.Header.Get("Origin")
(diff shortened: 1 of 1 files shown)


broker/lead.go
```diff
@@ -115,6 +109,41 @@ func lead(authToken models.AuthToken, policy *models.Policy) error {
 	return err
 }
 
+func checkIfPolicyIsLead(policy *models.Policy) error {
+	var recoveredPolicy models.Policy
+	log.Printf("[lead] found lead for existing policy %s", policy.Uid)
```
diogo-wopta: move log to line 129, after we successfully retrieved the policy from the DB

broker/lead.go
```diff
@@ -115,6 +109,41 @@ func lead(authToken models.AuthToken, policy *models.Policy) error {
...
+		return nil
+	}
+
+	if recoveredPolicy.Status != models.PolicyStatusPartnershipLead && recoveredPolicy.Status != models.PolicyStatusInitLead {
+		log.Printf("[lead] error policy %s is not a lead", policy.Uid)
+		return errors.New("policy is not a lead")
+	}
+
+	policy.CreationDate = recoveredPolicy.CreationDate
+	policy.Updated = recoveredPolicy.Updated
```
diogo-wopta: updated field can be removed as it will be set to now if the changes are successful

diogo-wopta (approved):

broker/lead.go
nahidalgo: Done in https://github.com/wopta/goworkspace/pull/121/commits/0d1cfbfbdb3a1bfc445c793fda5ffeef4f83cc4c

broker/lead.go
nahidalgo: Done in https://github.com/wopta/goworkspace/pull/121/commits/0d1cfbfbdb3a1bfc445c793fda5ffeef4f83cc4c

diogo-wopta (approved):

PR #130, opened 2023-10-25
HOTFIX: bug sumInsuredLimitOfIndemnity quote life v2

Yousef-Hammar:


quote/life.go
```diff
@@ -95,9 +95,17 @@ func Life(data models.Policy, channel string, networkNode *models.NetworkNode, w
...
+	case models.ProductV1:
+		log.Printf("[Life] product version %s", data.ProductVersion)
+		death, err := data.ExtractGuarantee(deathGuarantee)
+		lib.CheckError(err)
+		log.Println("[Life] setting sumInsuredLimitOfIndeminity")
+		calculateSumInsuredLimitOfIndemnity(data.Assets, death.Value.SumInsuredLimitOfIndemnity)
+		log.Println("[Life] setting guarantees duration")
+		calculateGuaranteeDuration(data.Assets, contractorAge, death.Value.Duration.Year)
+	case models.ProductV2:
+		log.Printf("[Life] product version %s", data.ProductVersion)
```
Yousef-Hammar: I made these changes because I realized that the previous code didn't meet the requirements for Life v2 E-commerce. The switch condition was based on channels, excluding E-commerce. Additionally, the old code was never deployed to production.

diogo-wopta: It seems correct that we should check for the version instead of the channel. The only thing that came up to my eyes is that the implementation of calculateSumInsuredLimitOfIndemnity is different from the calculateGuaranteeDuration, in the sense that one modifies directly the inputted assets and the other the guarantee inside the loop. Wouldn't want that the data won't be available outside the function since those do not return new assets and do not receive pointers.

PR #155, opened 2023-11-29
TG 318 321 | Network node proponent fields

diogo-wopta:


document/fpdf-utils.go
```diff
@@ -28,6 +28,13 @@ const (
 	proposal         = "PROPOSTA"
 )
 
+var productCompanyMap map[string]string = map[string]string{
+	models.LifeProduct:    "AXA FRANCE VIE S.A.",
+	models.GapProduct:     "Sogessur SA",
+	models.PmiProduct:     "Global Assistance",
+	models.PersonaProduct: "Global Assistance",
+}
```
Yousef-Hammar: Should we use ` company` as key for this map?

models/network-node.go
```diff
@@ -61,6 +62,8 @@ type AgencyNode struct {
 	RuiRegistration    time.Time             `json:"ruiRegistration" firestore:"ruiRegistration" bigquery:"-"`
 	BigRuiRegistration bigquery.NullDateTime `json:"-" firestore:"-" bigquery:"ruiRegistration"`
 	Skin               *Skin                 `json:"skin,omitempty" firestore:"skin,omitempty" bigquery:"-"`
+	Pec                string                `json:"pec,omitempty" firestore:"pec,omitempty" bigquery:"pec"`
+	Website            string                `json:"website,omitempty" firestore:"website,omitempty" bigquery:"website"`
```
Yousef-Hammar: do we need to save these fields is a column in bigquery tables?

diogo-wopta: I'm all for saving everything in big query to avoid missing fields for future queries/analysis

diogo-wopta: Deleted the fields from bigquery

Yousef-Hammar (approved):

PR #168, opened 2023-12-19
feat: script network transactions by manual payment

diogo-wopta:


diogo-wopta:
@Yousef-Hammar I'd merge this after the import release. It is a script we already used so it would be better to have it versioned in git. In any case if we need to use it again in the future we must always try them as 'dry-run' before hand.

PR #175, opened 2024-01-05
TG-385 | cancel and recreate transactions

diogo-wopta:


transaction/delete-transaction.go
```diff
@@ -0,0 +1,67 @@
...
+		return "", "", fmt.Errorf(errMessage)
+	}
+
+	err := DeleteTransaction(transaction, origin, "Cancellata manualmente")
+	if err != nil {
+		log.Printf("could not delete transaction '%s'", transactionUid)
+		return "", "", err
+	} else {
+		log.Printf("transaction '%s' successfully deleted", transactionUid)
+	}
```
diogo-wopta: unnecessary else block

payment/fabrick-refresh-pay-by-link.go
```diff
@@ -0,0 +1,256 @@
...
+	mail.SendMailPay(
+		policy,
+		mail.AddressAnna,
+		toAddress,
+		mail.Address{},
+		flowName,
+	)
+
+	return nil
+}
```
Yousef-Hammar, line 259: Should this be moved to `mail` domain?

payment/fabrick-refresh-pay-by-link.go
```diff
@@ -0,0 +1,256 @@
...
+			continue
+		}
+		scheduleDate, err := time.Parse(models.TimeDateOnly, tr.ScheduleDate)
+		if err != nil {
+			log.Printf("error parsing schedule date '%s' of transaction '%s': %s", tr.ScheduleDate, tr.Uid, err.Error())
+			return nil, nil, err
+		}
+		log.Printf("adding transcation with schedule date '%s' to be deleted and recreated", tr.ScheduleDate)
+		refreshScheduleDates = append(refreshScheduleDates, scheduleDate)
+		toBeDeletedTransactions = append(toBeDeletedTransactions, tr)
```
Yousef-Hammar: Add check on transaction.IsDeleted, if true do not append the transaction

payment/fabrick-refresh-pay-by-link.go
diogo-wopta: As we've seen on other domains, this step is usually done outside the mail domain by the current use case (ex.: broker). To move it to the mail domain would infer a change in all other uses cases. To be seen if that is the correct strategy when/if we'll have a sort of bpmn for the mail domain.

PR #193, opened 2024-01-31
TG-448 | Upload document error

diogo-wopta:


diogo-wopta:
Regardless of solving the upload for an empty file, the complete fix should also include a change to the Frontend as to not invoke this api without a base64 since the data for the document is not used in this api to update the policy object, that happens in the updatePolicy api that gets called aftwerwards.

Yousef-Hammar (approved):

PR #195, opened 2024-02-12
TG-461 | nodes import 

Yousef-Hammar:


network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+
+		pubSubClient, err := pubsub.NewClient(context.Background(), os.Getenv("GOOGLE_PROJECT_ID"))
+		if err != nil {
+			return "{}", nil, err
+		}
+		topic := pubSubClient.Topic("dataflow")
+		topic.Publish(context.Background(), &pubsub.Message{
+			Attributes: map[string]string{
+				"filename":       filename,
+				"invokerAddress": invokerAddress,
```
diogo-wopta, line 237: Let's add another attribute: 
"module": "in_network_node"
so we are sure of what module to run on dataflow, for now it's set at the trigger variables but that should not be the case

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+	// convert csv to bytes
+	data, err := base64.StdEncoding.DecodeString(req.Bytes)
+	if err != nil {
+		log.Printf("Error decoding file: %s", err.Error())
+		return "{}", nil, err
+	}
+
+	// load dataframe
+	df := lib.CsvToDataframe(data)
+	log.Printf("#rows: %02d #cols: %02d", df.Nrow(), df.Ncol())
```
diogo-wopta, line 118: should we error if the ncol is not what we expect? Since we access them by index, if a col is not present, does the function panic?

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+			DuplicatedNodes:           make([]string, 0),
+			InvalidConfigurationNodes: make([]string, 0),
+		},
+		ValidNodes: make([]string, 0),
+	}
+
+	// validate csv rows
+
+	for _, row := range df.Records()[1:] {
+		// normalize cells content if err add to skipped rows
```
diogo-wopta: update comment, the function does not return an error

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+			requiredFields = append(requiredFields, 30)
+			requiredFields = append(requiredFields, 31)
+		}
+	}
+
+	// check fiscalCode format
+	fiscalCodePattern := "^(?:[A-Z][AEIOU][AEIOUX]|[AEIOU]X{2}|[B-DF-HJ-NP-TV-Z]{2}[A-Z]){2}(?:[\\dLMNP-V]{2}(?:[" +
+		"A-EHLMPR-T](?:[04LQ][1-9MNP-V]|[15MR][\\dLMNP-V]|[26NS][0-8LMNP-U])|[DHPS][37PT][0L]|[ACELMRT][37PT][01LM]|[" +
+		"AC-EHLMPR-T][26NS][9V])|(?:[02468LNQSU][048LQU]|[13579MPRTV][26NS])B[26NS][9V])(?:[A-MZ][1-9MNP-V][\\dLMNP-V]" +
+		"{2}|[A-M][0L](?:[1-9MNP-V][\\dLMNP-V]|[0L][1-9MNP-V]))[A-Z]$"
```
diogo-wopta: can't we save it as a const outside the functions?

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+		// check if parent is present in nodesMap, if not skip
+		if reflect.ValueOf(parentNode).IsZero() {
+			log.Printf("Error processing node %s: parent node not found", nodeCode)
+			invalidConfigurationNodes = append(invalidConfigurationNodes, nodeCode)
+			continue
+		}
+
+		// check if parent is an agent in nodesMap, if not skip
+		if parentNode.Type == models.AgentNetworkNodeType {
+			log.Printf("Error processing node %s: agency can't have parent node of type agent", nodeCode)
```
diogo-wopta: we might error here even if the current node being verified is an agent, so the message is not exactly correct. maybe replace 'agency' with 'node'

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+				invalidConfigurationNodes = append(invalidConfigurationNodes, nodeCode)
+				continue
+			} else if !isMgaProponent && hasAnnex && designation == "" {
+				log.Printf("Error processing node %s: invalid node configuration for isMgaProponent = false", nodeCode)
+				invalidConfigurationNodes = append(invalidConfigurationNodes, nodeCode)
+				continue
+			}
+		} else if nodeType == models.AgentNetworkNodeType {
+			// check if fields for simplo are configured correctly
+			if isMgaProponent && (!hasAnnex || designation == "" || worksForUid == "" || (worksForUid != "__wopta__" && nodesMap[worksForUid].RuiSection != "E")) {
```
diogo-wopta: there is a constant for __wopta__

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+			continue
+		}
+
+		if nodeType == models.AgencyNetworkNodeType {
+			// check if fields for simplo are configured correctly
+			if worksForUid != "" {
+				log.Printf("Error processing node %s: not empty worksForUid", nodeCode)
+				invalidConfigurationNodes = append(invalidConfigurationNodes, nodeCode)
+				continue
+			}
```
diogo-wopta, line 468: Isn't there a case where it is empty and we use the node itself?

network/import-nodes.go
```diff
@@ -0,0 +1,522 @@
...
+
+		// check if parent is present in nodesMap, if not skip
+		if reflect.ValueOf(parentNode).IsZero() {
+			log.Printf("Error processing node %s: parent node not found", nodeCode)
+			invalidConfigurationNodes = append(invalidConfigurationNodes, nodeCode)
+			continue
+		}
+
+		// check if parent is an agent in nodesMap, if not skip
+		if parentNode.Type == models.AgentNetworkNodeType {
```
Yousef-Hammar, line 439: check also that an agency can't have another agency as parent

diogo-wopta (approved):

diogo-wopta (approved):

PR #200, opened 2024-02-21
NO-TG | check duplicated email import node+refactoring error response

Yousef-Hammar:


network/import-nodes.go
```diff
@@ -529,3 +561,50 @@ func writeCSVToBucket(outputRows [][]string, filename string) error {
...
+		log.Printf("Error getting invoker authToken: %s", err.Error())
+		return err
+	}
+	invokerAddress := authToken.Email
+	log.Printf("Invoker address: %s", invokerAddress)
+
+	pubSubClient, err := pubsub.NewClient(context.Background(), os.Getenv("GOOGLE_PROJECT_ID"))
+	if err != nil {
+		return err
+	}
```
diogo-wopta: add log for error

network/import-nodes.go
```diff
@@ -271,8 +274,11 @@ func buildNetworkNodesMap(dbNodes []models.NetworkNode) map[string]nodeInfo {
 			RuiSection:     ruiSection,
 		}
 
+		if !lib.SliceContains(emailsList, nn.Mail) {
+			emailsList = append(emailsList, nn.Mail)
+		}
```
diogo-wopta: Technically this control should not be necessary: if we are adding a node to the nodesMap, the email should be already unique, and later when we validate the rows we check if the email is present as to not add duplication. The 'problem' here is that we could potentially have a nodesMap with a number of keys different from the length of the emailsList.

And for the sake of duplication control when validating rows, it doesn't matter if it matches once or more times, if there is a match it errors.

network/import-nodes.go
```diff
@@ -345,84 +355,98 @@ func normalizeFields(row []string) []string {
...
+			log.Printf("Error processing node %s: missing required field at index %02d", nodeCode, index)
+			columnsError = append(columnsError, index)
 		}
 		_, err := time.Parse("02012006", fmt.Sprintf("%08s", row[index]))
 		if err != nil && lib.SliceContains(requiredFields, index) {
-			return fmt.Errorf("malformed date at index: %02d", index)
+			log.Printf("Error processing node %s: malformed date at index %02d", nodeCode, index)
+			columnsError = append(columnsError, index)
 		}
 	}
```
diogo-wopta: maybe rename index in 'colNumber' to avoid confusion with array index

diogo-wopta (approved):

PR #216, opened 2024-04-03
HOTFIX | policy queries

Yousef-Hammar:
Implemented join between policiesView and networkNodesView in order to be able to return also producer name in query result


policy/policy-info-bigquery-repository.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func buildPolicyInfoQuery(datasetID string, tableID string, queries []models.Query, limit int) (string, map[string]interface{}, error) {
+	var query bytes.Buffer
+	params := make(map[string]interface{})
+	fieldRegexp := regexp.MustCompile("^[a-zA-Z0-9.]*$")
+
+	query.WriteString(fmt.Sprintf("SELECT p.uid, p.name AS productName, p.codeCompany, CAST(p.proposalNumber AS INT64) AS proposalNumber, " +
+		"p.nameDesc,p.status, COALESCE(JSON_VALUE(p.data, '$.contractor.name'), '') || ' ' || " +
+		"COALESCE(JSON_VALUE(p.data, '$.contractor.surname'), '') AS contractor, " +
```
diogo-wopta: this also works for the PG policies?

policy/policy-info-bigquery-repository.go
```diff
@@ -0,0 +1,134 @@
...
+		"p.priceGross AS price, p.priceGrossMonthly AS priceGross, nn.name AS producer, p.producerCode, p.startDate, " +
```
diogo-wopta: shouldn't it be p.priceGrossMonthly AS priceMonthly ?

policy/policy-info-bigquery-repository.go
```diff
@@ -0,0 +1,134 @@
...
+				valueParameter := randomString(12)
+				params[valueParameter] = v
+				valuesParameters = append(valuesParameters, valueParameter)
+			}
+		} else {
+			// value parameter should be a random string of only letters
+			valueParameter := randomString(12)
+			params[valueParameter] = q.Value
+			valuesParameters = append(valuesParameters, valueParameter)
+		}
```
diogo-wopta: Can we refactor this? the only change between single and multiple values i the value to be used, but the structure is the same: generate key, add value to map, append key to keyList.

diogo-wopta (approved):

PR #233, opened 2024-05-14
TG-594 | Partnership WIN

diogo-wopta:
includes callback engine (TG-597)


diogo-wopta (commented):
Refactor auth functions, lots of repetitions
Review auth jwt parsing, use same strategy as partnership

diogo-wopta:
TODO:
- set win base path and auth with env vars
- save request, response and error

PR #239, opened 2024-06-17
TG-555 | renew mails

Yousef-Hammar:


mail/mail-policy.go
```diff
@@ -291,8 +285,39 @@ func SendMailProposal(policy models.Policy, from, to, cc Address, flowName strin
...
-		// LinkLabel:    "Leggi documentazione",
+	})
+}
+
+func SendMailRenewDraft(policy models.Policy, from, to, cc Address, flowName string, hasMandate bool) {
+	var bodyData BodyData
+
+	if flowName == models.RemittanceMgaFlow {
+		return
+	}
```
diogo-wopta: remove business logic from unit function

mail/mail-utils.go
```diff
@@ -74,6 +78,20 @@ func setPolicyReservedBodyData(policy models.Policy, bodyData *BodyData) {
 	}
 }
 
+func getPolicyRenewDraftBodyData(policy models.Policy, hasMandate bool) BodyData {
+	priceGross := policy.PriceGross
+	if policy.PaymentSplit == string(models.PaySplitMonthly) {
+		priceGross = policy.PriceGrossMonthly
```
diogo-wopta, line 85: my only concern here is for unformatted values like: 98,000000034
Should we use the data as string like:
```golang
value := policy.PriceGross
if policy.PaymenSplit == string(models.PaySplitMonthly) {
   value = polic.PriceGrossMonthly
}
priceGross := fmt.Sprintf("%.2f", value)
```
would need to update
```golang
struct BodyData {
   ...
   PriceGross string
}
```

diogo-wopta (approved):
Should do a go mod tidy to correctly put go-humanize as a direct dependency

diogo-wopta (approved):

PR #244, opened 2024-07-04
TG-658 | API renew policies

Yousef-Hammar:


policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+
+		{"startDateFrom": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"startDateTo": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"company": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"product": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"producerUid": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"status": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"payment": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
+		{"renewMonth": []string{"startDateFrom", "startDateTo", "company", "product", "producerUid", "reserved", "status", "payment", "renewMonth"}},
```
diogo-wopta: it seems to be missing the following entry:
```
{"reserved": []string{....}}
```

otherwise it should be removed from the allowedLists

policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	paramsWhereClause = map[string]string{
+		"codeCompany": "(codeCompany = @%s)",
+
+		"proposalNumber": "(proposalNumber = CAST(@%s AS INTEGER))",
```
diogo-wopta, line 39: Do we still need the CAST ?

policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+		if !lib.SliceContains(allowedParams, key) {
+			delete(params, key)
+		}
+	}
+	return params
+}
+
+func (qb *BigQueryQueryBuilder) processOrClauseParam(paramValue string) string {
+	whereClauses := make([]string, 0)
+	statusList := strings.Split(paramValue, ",")
```
diogo-wopta: consider renaming `statusList` into `paramsValueList` or something similar

policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (qb *BigQueryQueryBuilder) processProducerUidParam(paramKey, paramValue string, queryParams map[string]interface{}) string {
+	tmp := make([]string, 0)
+	for _, uid := range strings.Split(paramValue, ",") {
+		randomIdentifier := qb.randomGenerator()
+		queryParams[randomIdentifier] = uid
+		tmp = append(tmp, fmt.Sprintf("'@%s'", randomIdentifier))
+	}
+	return fmt.Sprintf(paramsWhereClause[paramKey], strings.Join(tmp, ", "))
+}
```
diogo-wopta: `paramKey` could be hardcoded since the function is used exclusively by `"producerUid"`

policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+			queryParams[randomIdentifier] = paramValue
+		}
+
+	}
+	return whereClauses, queryParams
+}
+
+func (qb *BigQueryQueryBuilder) BuildQuery(params map[string]string) (string, map[string]interface{}) {
+	const queryPrefix = "SELECT **tableAlias**.uid, **tableAlias**.name AS productName, " +
+		"**tableAlias**.codeCompany, CAST(**tableAlias**.proposalNumber AS INT64) AS proposalNumber, " +
```
diogo-wopta, line 166: in the`paramsWhereClause` map the CAST is into INTEGER and here to INT64

policy/query-builder/query-builder.go
```diff
@@ -0,0 +1,214 @@
...
+		"**tableAlias**.nameDesc,**tableAlias**.status, RTRIM(COALESCE(JSON_VALUE(**tableAlias**.data, " +
+		"'$.contractor.name'), '') || ' ' || " +
+		"COALESCE(JSON_VALUE(**tableAlias**.data, '$.contractor.surname'), '')) AS contractor, " +
```
diogo-wopta, line 169: for the contractor:
I'm not sure if we should do the `RTRIM` + `COALESCE` combo here. Since the response model from the api could potentially change, it would have unnecessary impact in the query.
An alternative would be to keep `contractorName` and `contractorSurname` base on the values of `$.data.contractor.name/surname`, and when marshaling the response do the concatenation.

policy/query-builder/query-builder_test.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/wopta/goworkspace/lib"
+	query_builder "github.com/wopta/goworkspace/policy/query-builder"
+)
+
+func TestQueryBuilder(t *testing.T) {
+	qb := query_builder.NewBigQueryQueryBuilder(lib.RenewPolicyViewCollection, "rp", func() string {
```
diogo-wopta: should use a fake name for the collection, ex.: `table-name`

policy/query-builder/query-builder_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+			if !strings.EqualFold(whereClauses, tc.want) {
+				t.Errorf("expected: %s, got: %s", tc.want, whereClauses)
+			}
+		})
+	}
+}
+
+func TestQueryBuilderFail(t *testing.T) {
+	qb := query_builder.NewBigQueryQueryBuilder(lib.RenewPolicyViewCollection, "rp", func() string {
```
diogo-wopta: same as above for the table name

diogo-wopta (commented):
Remember to include the .gcloudignore file

policy/query-builder/query-builder.go
Yousef-Hammar: we agreed to make this change later

diogo-wopta (approved):

PR #255, opened 2024-07-16
TG-672 | restore renew transaction

Yousef-Hammar:


transaction/restore-transaction.go
```diff
@@ -3,37 +3,64 @@ package transaction
 import (
 	"encoding/json"
 	"errors"
+	"log"
+	"net/http"
+	"strconv"
+
 	"github.com/go-chi/chi/v5"
+	plcRenew "github.com/wopta/goworkspace/policy/renew"
+	trxRenew "github.com/wopta/goworkspace/transaction/renew"
```
diogo-wopta: remove empty newline

transaction/restore-transaction.go
```diff
@@ -42,16 +69,24 @@ func RestoreTransactionFx(w http.ResponseWriter, r *http.Request) (string, inter
 		return "", nil, err
 	}
 
-	err = lib.SetFirestoreErr(models.TransactionsCollection, transaction.Uid, transaction)
+	err = saveTransactionToDb(collection, *transaction)
```
diogo-wopta: we have a `common.SaveTransactionsToDB` function. Should/could we use it?
Something like:
```
if err = common.SaveTransactionsToDB([]models.Transaction{*transaction}, collection); err != nil {
   ...
}
```

diogo-wopta (changes requested):

diogo-wopta (approved):

PR #256, opened 2024-07-16
TG-673 | manual payment renew transaction

Yousef-Hammar:


payment/manual/manual.go
```diff
@@ -215,14 +214,15 @@ func manualPayment(transaction *models.Transaction, origin string, payload *Manu
...
 		log.Printf("error saving transaction to firestore: %s", err.Error())
 		return err
 	}
 
-	transaction.BigQuerySave(origin)
-
-	return nil
+	transaction.BigQueryParse()
+	return lib.InsertRowsBigQuery(lib.WoptaDataset, collection, transaction)
 }
```
diogo-wopta: same as in the other PR. Should/could we use the `common.SaveTransactionsToDB` utils?

payment/manual/renew-manual.go
```diff
@@ -0,0 +1,107 @@
...
+
+	policy, err = plcRenew.GetRenewPolicyByUid(transaction.PolicyUid)
+	if err != nil {
+		return "", nil, err
+	}
+
+	isFirstTransactionAnnuity := policy.StartDate.AddDate(policy.Annuity, 0, 0) != transaction.EffectiveDate
+	if !isFirstTransactionAnnuity {
+		return "", nil, errors.New("cannot pay transaction that is not the first")
+	}
```
diogo-wopta, line 71: I'm not sure this is the business requirement but I agree that in this view we should be able to only act on the transaction that modifies also the draft, i.e. the first transaction of a given annuity. All other actions on transactions should be made in the centralized view of policy.

payment/manual/renew-manual.go
```diff
@@ -0,0 +1,107 @@
...
+
+	err = tr.CreateNetworkTransactions(&policy, transaction, networkNode, mgaProduct)
+	if err != nil {
+		return "", nil, err
+	}
+
+	err = saveTransaction(lib.RenewTransactionCollection, *transaction)
+	if err != nil {
+		return "", nil, fmt.Errorf(errPaymentFailed)
+	}
```
diogo-wopta: same as before for reusing `common.SaveTransactionsToDB`

diogo-wopta (approved):

PR #260, opened 2024-07-29
TG-682 | GetQuoterFile

Yousef-Hammar:


mga/get-quoter-file.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		log.Printf("Handler end -----------------------------------------------")
+		log.SetPrefix("")
+	}()
+
+	if authToken, err := lib.GetAuthTokenFromIdToken(r.Header.Get("Authorization")); err != nil || !authToken.IsNetworkNode {
+		log.Printf("error validating token")
+		err = errors.New("invalid token")
+		return "", nil, err
+	}
```
diogo-wopta: Maybe split error checking?
we are always overriding the value of err in case it is not nil..

diogo-wopta (approved):

PR #267, opened 2024-09-11
TG-736 | improve get products

fpittoni:
… finds some files outside /v[1-2]/ path

### DIFF
--- a/product/get-products.go
+++ b/product/get-products.go
@@ -2,14 +2,25 @@ package product
 
 import (
 	"encoding/json"
-	"log"
-	"os"
-	"sort"
-	"strings"
-
+	"fmt"
 	"github.com/wopta/goworkspace/lib"
 	"github.com/wopta/goworkspace/models"
+	"log"
+	"os"
+	"regexp"
+	"sort"
+	"strings"
+	"time"
 )
+
+func printSlice(stringSlice []string) int {
+	tot := 0
+	for _, v := range stringSlice {
+		fmt.Printf("%v\n", v)
+		tot++
+	}
+	return tot
+}
 
 func GetAllProductsByChannel(channel string) []models.ProductInfo {
 	log.Println("[GetAllProductsByChannel] function start -----------------------")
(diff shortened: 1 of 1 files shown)


product/get-products.go
```diff
@@ -2,15 +2,26 @@ package product
...
 	"log"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
-
-	"github.com/wopta/goworkspace/lib"
-	"github.com/wopta/goworkspace/models"
+	"time"
 )
```
diogo-wopta: Here we must set your GoLand to the same fmt that we use so we don't have unnecessary changes to import order.
It's alphabetical but the external modules are separated from the core ones.

product/get-products.go
```diff
@@ -136,5 +147,48 @@ func getProductsFileList() []string {
...
+	tot := printSlice(fileList)
+	log.Printf("There are %d elements", tot)
+	start := time.Now()
+	for _, s := range fileList {
+		if strings.Count(s, "/") == 3 {
+			filteredFileList = append(filteredFileList, s)
+		} else {
+			removedElements = append(removedElements, s)
+		}
+	}
```
diogo-wopta: Here you could use our lib.SliceFilter function that returns a new slice where all itens pass the given test.

product/get-products.go
```diff
@@ -136,5 +147,48 @@ func getProductsFileList() []string {
...
+	filteredFileList = nil
+	removedElements = nil
+	start = time.Now()
+	for _, s := range fileList {
+		if pattern.MatchString(s) {
+			filteredFileList = append(filteredFileList, s)
+		} else {
+			removedElements = append(removedElements, s)
+		}
+	}
```
diogo-wopta: Regex are powerful but also difficult. The given pattern has a problem, it would fail for a v10 because the character 0 is not allowed.
And it could also be double-edged, this check could pass for any file that has the version somewhere in the path, but it should be placed at a specific point.

product/get-products.go
```diff
@@ -136,5 +147,48 @@ func getProductsFileList() []string {
...
+		COUNT SEPARATORS
+	*/
+	var filteredFileList []string
+	var removedElements []string
+	log.Println("=== ORIGINAL LIST ===")
+	tot := printSlice(fileList)
+	log.Printf("There are %d elements", tot)
+	start := time.Now()
+	for _, s := range fileList {
+		if strings.Count(s, "/") == 3 {
```
diogo-wopta: one possible improvement would be to check that the number of '/' is bigger than 3, allowing for further nesting in the future.

diogo-wopta (commented):
@Yousef-Hammar I'm undecided here on what approach to go for:
1 (simpler) - just checking that the file is at the third level in the folder from products
2 (more complex) - checking for a certain pattern: starting with products/{name}/{version}, then having whatever content besides '/', and ending with a file extension (. + any word/digit).

fpittoni:
The first solution is the easier and faster (more or less ten time faster from my tests) but it forces to have a less flexible path structure.
The second is slower (which make sense: regexps are slow) but  enforces a Vn check (and Diogo is right, it doesn't work with v10 or v0 but the fix is not a big deal).
Probably the best trade off is to check that the number of "/" is 3 or more, as Diogo suggests, allowing for deeper nesting, which is why the regexp solution was implemented in the first place.

Let me know what you want me to do: I can write the code and open a PR in a couple of minutes :-)

Yousef-Hammar:
> @Yousef-Hammar I'm undecided here on what approach to go for: 1 (simpler) - just checking that the file is at the third level in the folder from products 2 (more complex) - checking for a certain pattern: starting with products/{name}/{version}, then having whatever content besides '/', and ending with a file extension (. + any word/digit).

I would go with the simpler solution and delay the decision regarding a naming convention to the future

diogo-wopta:
> @Yousef-Hammar I'm undecided here on what approach to go for: 1 (simpler) - just checking that the file is at the third level in the folder from products 2 (more complex) - checking for a certain pattern: starting with products/{name}/{version}, then having whatever content besides '/', and ending with a file extension (. + any word/digit).

@fpittoni talking with @Yousef-Hammar we decided on the first approach.

EDIT: sorry for the repeated messages. I had not seen that Filippo and Yousef had already added their comments.

product/get-products.go
Yousef-Hammar: image

@fpittoni these are my settings for import:

GoLand -> Settings -> Editor -> Code Style -> Go

diogo-wopta (approved):

PR #269, opened 2024-09-13
TG-755 | Persona Reserved BMI check

fpittoni:


reserved/persona.go
```diff
@@ -0,0 +1,51 @@
...
+func personaReserved(policy *models.Policy) (bool, *models.ReservedInfo) {
+	log.Println("[personaReserved]")
+
+	reservedInfo := &models.ReservedInfo{
+		Reasons:       make([]string, 0),
+		RequiredExams: make([]string, 0),
+	}
+
+	isReserved := false
+	reason := "BMI index out of range"
```
diogo-wopta: Change it into a constant so we know that it's value can't/shouldn't be changed. For now it can be inside this function as it's not used elsewhere.

reserved/persona.go
```diff
@@ -0,0 +1,51 @@
...
+
+	if len(policy.Assets) == 0 || policy.Assets[0].Person == nil || policy.Assets[0].Person.Weight == 0 || policy.Assets[0].Person.Height == 0 {
+		return isReserved, reservedInfo
+	}
```
diogo-wopta: we could enhance this control to make it more legible. One option is to make to variables that hold different kind of information as: policyHasAsset, and personIsInitialized (the naming is just an example, better ones would be an improvement)

reserved/persona.go
```diff
@@ -0,0 +1,51 @@
...
+	w := float64(weight)
+	h := float64(height)
+	bmi := w / math.Pow(h/100, 2)
+	isOutOfRange := false
+	if (bmi-16) < epsilon || (bmi-30) > epsilon {
+		isOutOfRange = true
+	}
+
+	return bmi, isOutOfRange
+}
```
diogo-wopta: Really interesting approach that raises the question of what is our precision (how many floating places) and how should we round the values. We'll raise the question to business and then update the code accordingly.

reserved/persona_test.go
```diff
@@ -0,0 +1,56 @@
...
+			expected: true, // bmi = 30,06792
+		},
+	}
+
+	for i, test := range tests {
+		bmi, isOutOfRange := checkOutOfRangeBMI(test.weight, test.height)
+		if isOutOfRange != test.expected {
+			t.Errorf("BMI check failed for test n° %d: result is %t because bmi is %f", i+1, isOutOfRange, bmi)
+		}
+	}
```
diogo-wopta: Here we could use a 'expected != got' structure so something like:
```golang
_, got := checkOutOfRangeBMI(test.weight, test.height)
if test.expected != got {
    t.Errorf("error in test n° %d: expected %t, got %t", i+1, test.expected, got)
}
```

Another improvement would be to name the tests so that we can find them easily, rather than using the index

diogo-wopta:
I'll talk to Luca but we must fix the error on companydata before merging.

reserved/persona_test.go
Yousef-Hammar: Another approach would be to give to each test case a name. This name should be a small description of that specific test case. 

```golang
_, got := checkOutOfRangeBMI(test.weight, test.height)
if test.expected != got {
    t.Errorf("error in test %s %d: expected %t, got %t", test.name, test.expected, got)
}
```

diogo-wopta:
We've got an update on an email, the upper limit is **40**.
We'll try to get hold of the external implementation for the product to check how they handled the rounding/precision, since that version was authorized by the company.

reserved/persona.go
```diff
@@ -0,0 +1,51 @@
...
+		Numbers have no exact representation in binary floating point.
+		So instead of comparing values, we compare their difference against a very small
+		constant (called epsilon here)
+	*/
+	const epsilon = 1e-9
+	w := float64(weight)
+	h := float64(height)
+	bmi := w / math.Pow(h/100, 2)
+	isOutOfRange := false
+	if (bmi-16) < epsilon || (bmi-30) > epsilon {
```
diogo-wopta: extract the bmi limits (16 and 30 in this version) into two constants so that we can update them easily.

reserved/persona.go
```diff
@@ -16,9 +16,14 @@ func personaReserved(policy *models.Policy) (bool, *models.ReservedInfo) {
 	}
 
 	isReserved := false
-	reason := "BMI index out of range"
+	const reason = "BMI index out of range"
```
diogo-wopta: move the declaration up to be the first after the initial log

reserved/persona.go
```diff
@@ -16,9 +16,14 @@ func personaReserved(policy *models.Policy) (bool, *models.ReservedInfo) {
...
 
-	if len(policy.Assets) == 0 || policy.Assets[0].Person == nil || policy.Assets[0].Person.Weight == 0 || policy.Assets[0].Person.Height == 0 {
+	policyHasNoAsset := len(policy.Assets) == 0
+	assetHasNoPerson := policy.Assets[0].Person == nil
```
diogo-wopta: join these two as a single variable:
`{variableName} := len(policy.Assets) == 0 || policy.Assets[0].Person == nil`

reserved/persona.go
```diff
@@ -16,9 +16,14 @@ func personaReserved(policy *models.Policy) (bool, *models.ReservedInfo) {
...
+	personHasNoWeight := policy.Assets[0].Person.Weight == 0
+	personHasNoHeight := policy.Assets[0].Person.Height == 0
```
diogo-wopta: join these two as a single variable:
`{variableName} := policy.Assets[0].Person.Weight == 0 || policy.Assets[0].Person.Height == 0`

reserved/persona_test.go
```diff
@@ -3,54 +3,62 @@ package reserved
...
 		},
 	}
 
 	for i, test := range tests {
-		bmi, isOutOfRange := checkOutOfRangeBMI(test.weight, test.height)
-		if isOutOfRange != test.expected {
-			t.Errorf("BMI check failed for test n° %d: result is %t because bmi is %f", i+1, isOutOfRange, bmi)
+		_, got := checkOutOfRangeBMI(test.weight, test.height)
+		if test.expected != got {
+			t.Errorf("error in test %s %d: expected %t, got %t", test.description, i+1, test.expected, got)
```
diogo-wopta: Remove the index from the log, it's not necessary having the description.

diogo-wopta (approved):

diogo-wopta:
@fpittoni I've updated `master` with the fix for the action. Could you merge it into your branch and update the PR so we can re-run the tests?

PR #271, opened 2024-09-17
TG-486 | test for unique node code

fpittoni:
This PR has been opened in draft mode because I'm not so confident this is the best way to fix the issue: checking against an error to test if a node is already present is at least ambiguous. Yesterday I chose this solution to avoid code duplication but probably there are better ways (and I'd love to hear some good advice)

### DIFF
--- a/mga/create-network-node.go
+++ b/mga/create-network-node.go
@@ -35,7 +35,7 @@ func CreateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
 	// TODO: check node.Type in warrant.AllowedTypes
 	// TODO: check unique companyCode for company
 	if err := network.TestNetworkNodeUniqueness(inputNode.Code); err != nil {
-		log.Println(err)
+		log.Printf("error validating node code: %s", err)
 		return "", "", err
 	}
 
--- a/mga/create-network-node.go
+++ b/mga/create-network-node.go
@@ -2,6 +2,7 @@ package mga
 
 import (
 	"encoding/json"
+	"errors"
 	"io"
 	"log"
 	"net/http"
@@ -33,8 +34,16 @@ func CreateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
 		return "", "", err
 	}
 	// TODO: check node.Type in warrant.AllowedTypes
-	// TODO: check unique node.Code
 	// TODO: check unique companyCode for company
+	_, err = network.GetNetworkNodeByCode(inputNode.Code)
+	if !errors.Is(err, network.ErrNodeNotFound) {
+		// If err is *not* ErrNodeNotFound, options are:
+		// - an error raised hitting the db
+		// - the node code is already present
+		// In both cases we must abort
+		log.Printf("error checking node code: %s", err.Error())
+		return "", "", err
+	}
 
 	log.Println("creating network node into Firestore...")
 
--- a/mga/create-network-node.go
+++ b/mga/create-network-node.go
@@ -2,7 +2,6 @@ package mga
 
 import (
 	"encoding/json"
-	"errors"
 	"io"
 	"log"
 	"net/http"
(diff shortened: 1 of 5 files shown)


mga/update-network-node.go
```diff
@@ -32,6 +33,16 @@ func UpdateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
...
 
+	_, err = network.GetNetworkNodeByCode(inputNode.Code)
+	if !errors.Is(err, network.ErrNodeNotFound) {
+		// If err is *not* ErrNodeNotFound, options are:
+		// - an error raised hitting the db
+		// - the node code is already present
+		// In both cases we must abort
+		log.Printf("error checking node code: %s", err.Error())
+		return "", "", err
+	}
```
diogo-wopta: Even though a `nil` error gets catched by the control flow, we would have a panic here for the log line since `nil` does not have an `Error()` method. Regardless, the return response does not suffice in this case because we would return a `statusCode 200` with an empty body to the FE. In this case we must ensure that we return _some_ kind of error.

It seems we need to split the flow:
- error `nil` => node found so we exit with another type of error
- error not `nil` && not `ErrNodeNotFound` => we got some kind of unexpected error so we exit with that error

mga/create-network-node.go
```diff
@@ -33,8 +34,16 @@ func CreateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
...
 	// TODO: check unique companyCode for company
+	_, err = network.GetNetworkNodeByCode(inputNode.Code)
+	if !errors.Is(err, network.ErrNodeNotFound) {
+		// If err is *not* ErrNodeNotFound, options are:
+		// - an error raised hitting the db
+		// - the node code is already present
+		// In both cases we must abort
+		log.Printf("error checking node code: %s", err.Error())
+		return "", "", err
+	}
```
diogo-wopta: See the comment above for specs. Consider extracting the whole flow into a wrapper function since the same change we'll be reflected here. It could be something like:
```golang
if err := wrapperFunction(inputNode.Code); err != nil {
    log.Println(err)
    return "", "", err
}
```

diogo-wopta (changes requested):
On the next update, rename the PR with the name of the ticket and a short message:
`TG-XXX | Lorem Ipsum`

mga/create-network-node.go
```diff
@@ -33,8 +33,11 @@ func CreateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
 		return "", "", err
 	}
 	// TODO: check node.Type in warrant.AllowedTypes
-	// TODO: check unique node.Code
 	// TODO: check unique companyCode for company
+	if err := network.TestNetworkNodeUniqueness(inputNode.Code); err != nil {
+		log.Println(err)
```
diogo-wopta: Enhance the log message:
```golang
log.Printf("error validating node code: %s", err)
```

mga/update-network-node.go
```diff
@@ -32,6 +32,11 @@ func UpdateNetworkNodeFx(w http.ResponseWriter, r *http.Request) (string, interf
 		return "", "", err
 	}
 
+	if err := network.TestNetworkNodeUniqueness(inputNode.Code); err != nil {
+		log.Println(err)
```
diogo-wopta: Enhance the log message:
```golang
log.Printf("error validating node code: %s", err)
```

network/network_node_repository.go
```diff
@@ -413,32 +414,48 @@ func addWorksForUid(originalNode, inputNode *models.NetworkNode) error {
...
+
+func TestNetworkNodeUniqueness(nodeCode string) error {
+	_, err := GetNetworkNodeByCode(nodeCode)
+	// If err == nil, node found
+	if err == nil {
+		return errors.New("node already exists")
+	}
+	if !errors.Is(err, ErrNodeNotFound) {
+		// If err is *not* ErrNodeNotFound, an error raised hitting the db
+		log.Printf("error checking node code: %s", err.Error())
```
diogo-wopta: Let's try to enforce a standard here (if @Yousef-Hammar agrees):
_whenever we can, let's leave the logs to the caller_ -> in this case it will be the two API handlers.

network/network_node_repository.go
```diff
@@ -413,32 +414,48 @@ func addWorksForUid(originalNode, inputNode *models.NetworkNode) error {
...
+		return nil, err
 	}
 
-	return node
+	return node, nil
+}
+
+func TestNetworkNodeUniqueness(nodeCode string) error {
+	_, err := GetNetworkNodeByCode(nodeCode)
+	// If err == nil, node found
```
diogo-wopta: remove comment

network/network_node_repository.go
```diff
@@ -413,32 +414,48 @@ func addWorksForUid(originalNode, inputNode *models.NetworkNode) error {
...
+	if err == nil {
+		return errors.New("node already exists")
+	}
+	if !errors.Is(err, ErrNodeNotFound) {
+		// If err is *not* ErrNodeNotFound, an error raised hitting the db
```
diogo-wopta: remove comment

diogo-wopta (approved):

PR #273, opened 2024-09-20
NO-TG | fix missing watermark simplo persona

Yousef-Hammar:


document/contract-persona-global-v1.go
```diff
@@ -150,6 +150,10 @@ func personaMainHeaderV1(pdf *fpdf.Fpdf, policy *models.Policy, networkNode *mod
 		pdf.SetXY(-95, pdf.GetY()+3)
 		pdf.MultiCell(0, 3.5, contractorInfo, "", "", false)
 		pdf.Ln(5)
+
+		if isProposal {
```
diogo-wopta, line 154: do we need to check `isProposal`? Shouldn't this be only for the contract?

diogo-wopta (approved):

PR #276, opened 2024-10-03
wip: send email 10 days before policy renew

fpittoni:
The new endpoint can be triggered by a POST on {{baseUrl}}/renew/v1/rename-me with a payload like:
{
    "date": "2025-09-22",
    "days_before_renew": "10"
}

As you can notice, the endpoint has been temporary called "rename-me" and I'm waiting for a suitable name

### DIFF
--- a/renew/mail.go
+++ b/renew/mail.go
@@ -70,9 +70,9 @@ func RenewMailFx(w http.ResponseWriter, r *http.Request) (string, interface{}, e
 	for _, policy := range policies {
 		from := mail.AddressAnna
 		to := mail.GetContractorEmail(&policy)
-		//flowName := models.ECommerceFlow
+		flowName := models.ECommerceFlow
 		log.Printf("Sending email from %s to %s", from, to)
-		//mail.SendMailRenewDraft(policy, from, to, mail.Address{}, flowName, policy.HasMandate)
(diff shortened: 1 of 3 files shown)


renew/mail.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	targetDate := date.AddDate(0, 0, daysBeforeRenew)
+	log.Printf("Date is: %v; targetDate is: %v", date, targetDate)
+
+	params["isRenewable"] = true
+	params["channel"] = models.ECommerceChannel
+	params["targetYear"] = int64(targetDate.Year())
+	params["targetMonth"] = int64(targetDate.Month())
+	params["targetDay"] = int64(targetDate.Day())
```
Yousef-Hammar: Here you should add also:

- `isDeleted = false`
- `isPay = false`

So that we are sure first to send email only to people for which renew is in progress and payment has not been made. It could happen that a node from the network contacts the customer before -10 days and proceed with the payment. In this case we don't want to send an email saying you can pay .....

renew/mail.go
Yousef-Hammar, line 1: Maybe it could be worth it to separate business logic from the http handler in a separate function within the same file

diogo-wopta:
@fpittoni should we close this draft? It seems it is considered also in #278

PR #278, opened 2024-10-09
TG-679 | renew draft notice and error handling sendMail function

fpittoni:


lib/file-utils.go
```diff
@@ -280,6 +304,26 @@ func GetFilesByEnv(file string) []byte {
 	return res1
 }
 
+func GetFilesByEnvV2(file string) ([]byte, error) {
+	var (
+		res1 []byte
```
diogo-wopta: rename me

mail/mail-policy.go
```diff
@@ -318,3 +318,48 @@ func SendMailRenewDraft(policy models.Policy, from, to, cc Address, flowName str
 		IsApp:       true,
 	})
 }
+
+func SendMailRenewDraftV2(policy models.Policy, from, to, cc Address, flowName string, hasMandate bool) error {
```
diogo-wopta: if I'm not mistaken, policy should have a `HasMandate` flag now so it shouldn't be necessary to pass it as an extra argument

mail/mail-policy.go
```diff
@@ -318,3 +318,48 @@ func SendMailRenewDraft(policy models.Policy, from, to, cc Address, flowName str
...
+		SubTitle:    subtitle,
+		Subject:     subject,
+		IsHtml:      true,
+		IsApp:       true,
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
diogo-wopta: since we do not do extra work with the error, we could return directly the result from SendMailV2
`return SendMailV2(...)`

mail/send.go
```diff
@@ -271,3 +271,200 @@ func SendMail(obj MailRequest) {
...
+	)
+
+	switch os.Getenv("env") {
+	case "local":
+		file, err = os.ReadFile("../function-data/dev/mail/mail_template.html")
+	case "dev":
+		file, err = lib.GetFromStorageV2("function-data", "mail/mail_template.html", "")
+	case "prod":
+		file, err = lib.GetFromStorageV2("core-350507-function-data", "mail/mail_template.html", "")
+	}
```
diogo-wopta: missing a check for `err != nil` after the switch for an early escape before parsing the template

renew/mail.go
```diff
@@ -0,0 +1,150 @@
...
+	var (
+		err             error
+		date            = time.Now().UTC()
+		daysBeforeRenew = 10
+		req             RenewReq
+	)
+
+	log.SetPrefix("[RenewMailFx] ")
+	defer func() {
+		collectionPrefix = ""
```
diogo-wopta: are we using this variable?

renew/mail.go
```diff
@@ -0,0 +1,150 @@
...
+		log.Printf("error getting renew policies: %s", err.Error())
+		return "", nil, err
+	}
+
+	for _, policy := range policies {
+		from := mail.AddressAnna
+		to := mail.GetContractorEmail(&policy)
+		flowName := models.ECommerceFlow
+		log.Printf("Sending email from %s to %s", from, to)
+		//err = writeMailError(policy.Uid, to.Name, to.Address, lib.GetBigQueryNullDateTime(time.Now().UTC()), "test error")
```
diogo-wopta: remove me

renew/mail.go
```diff
@@ -0,0 +1,150 @@
...
+		err = mail.SendMailRenewDraftV2(policy, from, to, mail.Address{}, flowName, policy.HasMandate)
```
diogo-wopta: as commented on the mail part, remove the last argument since we pass the whole policy into the function

renew/mail.go
```diff
@@ -0,0 +1,150 @@
...
+	params["targetYear"] = int64(targetDate.Year())
+	params["targetMonth"] = int64(targetDate.Month())
+	params["targetDay"] = int64(targetDate.Day())
+
+	query.WriteString(fmt.Sprintf("SELECT * FROM `%s.%s` WHERE", lib.WoptaDataset, lib.RenewPolicyViewCollection))
+	query.WriteString(" isRenewable = @isRenewable")
+	query.WriteString(" AND isDeleted = @isDeleted")
+	query.WriteString(" AND isPay = @isPay")
+	query.WriteString(" AND channel = @channel")
+	query.WriteString(" AND EXTRACT(YEAR FROM RenewDate) = @targetYear")
```
diogo-wopta: check PR #283 since we might have the same problem here.

renew/mail.go
```diff
@@ -0,0 +1,150 @@
...
+		var tmpPolicy models.Policy
+		err = json.Unmarshal([]byte(p.Data), &tmpPolicy)
+		return tmpPolicy
+	})
+
+	return policies, nil
+}
+
+func writeMailError(policyId string, name string, address string, date bigquery.NullDateTime, message string) error {
+
```
diogo-wopta: remove empty line

renew/renew.go
```diff
@@ -21,6 +21,12 @@ var routes []lib.Route = []lib.Route{
 		Handler: lib.ResponseLoggerWrapper(PromoteFx),
 		Roles:   []string{},
 	},
+	{
+		Route:   "/v1/rename-me",
```
diogo-wopta: we could use `renew-notice` for now while we don't come up with a better name.

mail/send.go
```diff
@@ -27,6 +29,14 @@ type loginAuth struct {
...
 }
 
+type MailReport struct {
+	Policy       string                `bigquery:"policyUid"`
+	Name         string                `bigquery:"name"`
+	Email        string                `bigquery:"email"`
+	CreationDate bigquery.NullDateTime `bigquery:"creationDate"`
+	MailError    string                `bigquery:"mailError"`
+}
+
```
diogo-wopta: Move me to mail/models.go

Cleanup key names. Since we are doing something from scratch lets make them all coincide (Name in Go with name in BigQuery). If needed we can update the schema, not a big deal.

Also make it so that the use is clear. In the example above it doesn't seem so:
```golang
report := MailReport{policy, sender, recipients, date, message}
```

diogo-wopta (changes requested):
Afterwards, open the PR since it is still a Draft.

renew/renew.go
Yousef-Hammar: What about `/renew/notice` so that in the future we could follow a folder structure path pattern for different kind of emails regarding renew process

lib/constants.go
```diff
@@ -52,4 +52,5 @@ const (
 	RenewPolicyCollection        string = "renewPolicy"            // firestore and bigquery
 	RenewPolicyViewCollection    string = "renewPolicyView"        // bigquery
 	RenewTransactionCollection   string = "renewTransactions"      // firestore and bigquery
+	MailReportCollection         string = "mail-report"            // only for bigquery
```
Yousef-Hammar: @diogo-wopta here I would renew the table into `mailReport` to set a common naming rule

lib/file-utils.go
```diff
@@ -102,6 +102,30 @@ func GetFromStorage(bucket string, file string, keyPath string) []byte {
...
+	if err != nil {
+		return nil, err
+	}
+	err = rc.Close()
+	if err != nil {
+		return nil, err
+	}
+	return slurp, err
+}
+
```
Yousef-Hammar: What's the difference between this and `GetFromStorageErr` function?

mail/send.go
```diff
@@ -110,28 +120,34 @@ func getContentType(ext string) string {
...
 	case "local":
-		file = lib.ErrorByte(os.ReadFile("../function-data/dev/mail/mail_template.html"))
+		file, err = os.ReadFile("../function-data/dev/mail/mail_template.html")
 	case "dev":
-		file = lib.GetFromStorage("function-data", "mail/mail_template.html", "")
+		file, err = lib.GetFromStorageV2("function-data", "mail/mail_template.html", "")
 	case "prod":
-		file = lib.GetFromStorage("core-350507-function-data", "mail/mail_template.html", "")
+		file, err = lib.GetFromStorageV2("core-350507-function-data", "mail/mail_template.html", "")
+	}
```
Yousef-Hammar: use `GetFilesByEnvV2`

lib/file-utils.go
fpittoni: Honestly, when I started refactoring the email function, I didn't realised there was `GetFromStorageErr`.
IMHO `GetFromStorageV2` has a better error handling so I would be inclined to keep this, at least for this PR. But if you decide otherwise the fix is a one minute work.

lib/constants.go
fpittoni: Personally I agree. So I'll go with the renaming since the db will have to be modified anyway.

mail/send.go
fpittoni: The new struct is: 

```
type MailReport struct {
	Policy         string                `bigquery:"policyUid"`
	SenderName     string                `bigquery:"senderName"`
	RecipientEmail string                `bigquery:"recipientEmail"`
	CreationDate   bigquery.NullDateTime `bigquery:"creationDate"`
	MailError      string                `bigquery:"mailError"`
}
```

@diogo-wopta update/recreate bigquery table keeping in mind that now 
`MailReportCollection         string = "mailReport"`

diogo-wopta: @fpittoni table created on bigquery

renew/renew.go
diogo-wopta: @Yousef-Hammar the thing is that we are already at the `renew` root i.e.: the path currently is `host/renew/v1/renew-notice`. What we could do would be further split the path as so -> `host/renew/v1/notice/e-commerce` since in the other task #246 the goal is to send the notice for the network, and a possible path would be `host/renew/v1/notice/network`. What do you think?

lib/constants.go
diogo-wopta: table recreated on bigquery with the new name

lib/file-utils.go
Yousef-Hammar: Talking with Diogo, we have decided to remove  existent `GetFromStorageErr` and rename your `GetFromStorageV2` into `GetFromStorageErr`

Yousef-Hammar (changes requested):

mail/send.go
fpittoni: I tested it and it works. Don't forget to delete the old mail-report table.

mail/send.go
diogo-wopta, line 1: The function `lib.GetFilesByEnv2` should replace the whole switch statement, not the inner cases. That switch is already present in the aforementioned function.

diogo-wopta (changes requested):

diogo-wopta (approved):
Approved. If possible update only the name of the PR so that it starts with the taiga ticket: TG-679 | ...

Yousef-Hammar (approved):

diogo-wopta:
> Approved. If possible update only the name of the PR so that it starts with the taiga ticket: TG-679 | ...

Fatto da me stesso. @fpittoni Tutto pronto, mergio nel pomeriggio.

PR #281, opened 2024-10-18
NO-TG | fix persona

Yousef-Hammar:


diogo-wopta (changes requested):
This probably won't compile. Must address the uses of Enterprise Address as a string in other parts.
Must also change *Address in Building in the same file.
Files to update:
- companydata/pmi-global-emit.go
- document/widget.go
- models/wise-asset.go
- quote/combined-qbe.go

diogo-wopta (approved):

PR #283, opened 2024-10-21
NO-TG | Fix renew draft consider current year

diogo-wopta:
depending on the renew offset, the policy might be worked on the same year it is emitted. The fix aims to include also those policies avoiding an error at the end of the year that policies from the beginning of the following year wouldnt be considered


diogo-wopta:
@Yousef-Hammar whenever you can, let's merge this fix into the bundle that is going in dev before the tests so that we can also verify the start/end of year scenarios.

The only thing it came up to me is that there is a branch NO-TG_renew-darft-year-change with a different fix, supposedly to the same problem.. can you check it?

Yousef-Hammar (approved):

PR #286, opened 2024-10-30
TG-823 | enable request approval callback

Yousef-Hammar:
- Enabled RequestApproval event callback
- Disabled Signed, Approved, Rejected events callback


diogo-wopta (approved):
You may test it with a token following the instructions

PR #289, opened 2024-11-04
ISSUE-797 | cancel fabrick transaction without providerId

Yousef-Hammar:


payment/delete-transaction.go
```diff
@@ -43,20 +53,28 @@ func DeleteTransactionFx(w http.ResponseWriter, r *http.Request) (string, interf
...
+			return "", nil, fmt.Errorf("transaction '%s' not found", uid)
+		}
+		policy, err = renew.GetRenewPolicyByUid(transaction.PolicyUid)
 	}
-
-	if transaction == nil {
-		log.Printf("transaction '%s' not found", uid)
-		return "", nil, fmt.Errorf("transaction '%s' not found", uid)
+	if err != nil {
+		return "", nil, err
```
diogo-wopta, line 72: here the error should always be a 'policy not found'. Shall we log it as in the other error handlings?

diogo-wopta (approved):

PR #295, opened 2024-11-11
TG-855 | Manual payment for network

Yousef-Hammar:


models/constants.go
```diff
@@ -187,8 +187,14 @@ const (
 	PayMethodRemittance = "remittance"
 )
 
-func GetAllPaymentMethods() []string {
-	return []string{PayMethodCard, PayMethodTransfer, PayMethodSdd}
+func GetAllPaymentMethods(role string) []string {
```
diogo-wopta: Rename me into something more explicit, maybe `GetAvailableMethodsByRole` ?

diogo-wopta (commented):
Regardless of the name refactoring, it seems good to me.

models/constants.go
Yousef-Hammar: I thought having `role` as parameter was enough but I have no problem in changing function name as you suggested

diogo-wopta (approved):

PR #296, opened 2024-11-12
NO-TG | GetNetworkNode API

Yousef-Hammar:


mga/get-network-node-by-uid.go
```diff
@@ -10,23 +11,44 @@ import (
...
+	var (
+		err error
+	)
+
+	defer func() {
+		if err != nil {
+			log.Printf("error: %+v", err.Error())
+		}
+		log.SetPrefix("")
+		log.Println("Handler end -------------------------------------------------")
```
diogo-wopta, line 22: Empty the prefix only after logging Handler end

diogo-wopta (approved):

PR #298, opened 2024-11-13
NO-TG | fix manual payment

Yousef-Hammar:
In this PR I've added missing checks on policy producer and flow of authenticated user required to open this operation also to network nodes

### DIFF
--- a/payment/manual/manual.go
+++ b/payment/manual/manual.go
@@ -1,6 +1,7 @@
 package manual
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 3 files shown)


payment/manual/manual.go
```diff
@@ -40,14 +42,19 @@ func ManualPaymentFx(w http.ResponseWriter, r *http.Request) (string, interface{
 	)
 
 	log.SetPrefix("[ManualPaymentFx] ")
-	defer log.SetPrefix("")
-
+	defer func() {
+		if err != nil {
+			log.Printf("error: %s", err.Error())
+		}
+		defer r.Body.Close()
```
diogo-wopta: we are already inside a defer, the second defer doesn't seem necessary

payment/manual/manual.go
```diff
@@ -60,36 +67,61 @@ func ManualPaymentFx(w http.ResponseWriter, r *http.Request) (string, interface{
...
 
 	isMethodAllowed := lib.SliceContains[string](methods, payload.PaymentMethod)
 	if !isMethodAllowed {
-		log.Printf("ERROR %s", errPaymentMethodNotAllowed)
-		return "", nil, fmt.Errorf(errPaymentMethodNotAllowed)
+		err = fmt.Errorf("ERROR %s", errPaymentMethodNotAllowed)
+		return "", nil, err
 	}
 
 	origin := r.Header.Get("Origin")
```
diogo-wopta: can we delete this `origin` without a fuss? It shouldn't matter anymore

payment/manual/manual.go
```diff
@@ -60,36 +67,61 @@ func ManualPaymentFx(w http.ResponseWriter, r *http.Request) (string, interface{
...
+	warrant = networkNode.GetWarrant()
+	ccAddress = mail.GetNetworkNodeEmail(networkNode)
+	flowName, _ = policy.GetFlow(networkNode, warrant)
+	log.Printf("flowName '%s'", flowName)
+
+	isOnlineProducer := authToken.UserID != policy.ProducerUid || flowName != models.RemittanceMgaFlow
+	if authToken.IsNetworkNode && isOnlineProducer {
+		err = errors.New("cannot access to transaction")
+		return "", nil, err
+	}
```
diogo-wopta: I know that this might be a little obnoxious, but what if we controlled like this?
```golang
canUserAccessTransaction := authToken.Role == models.UserRoleAdmin || (authToken.IsNetworkNode && authToken.UserID == policy.ProducerUid && flowName == models.RemittanceMgaFlow)
if !canUserAccessTransaction {
        err = errors.New("user cannot access transaction")
        return "", nil, err
}
```

payment/manual/manual.go
```diff
@@ -60,36 +67,61 @@ func ManualPaymentFx(w http.ResponseWriter, r *http.Request) (string, interface{
...
 
 	if transaction.IsPay {
-		log.Printf("ERROR %s", errTransactionPaid)
-		return "", nil, fmt.Errorf(errTransactionPaid)
+		err = fmt.Errorf(errTransactionPaid)
```
diogo-wopta: `err = errTransactionPaid` should suffice.
The same goes for the errors below.

EDIT: just noticed that the err constants are not error but strings. Keep it the way it is or refactor them into errors, your call.
```golang
var (
        errTransactionPaid = errors.New("input message here")
        ...
)
```

payment/manual/renew-manual.go
```diff
@@ -84,6 +83,21 @@ func RenewManualPaymentFx(w http.ResponseWriter, r *http.Request) (string, inter
...
+	}
+	warrant := networkNode.GetWarrant()
+	flowName, _ := policy.GetFlow(networkNode, warrant)
+	log.Printf("flowName '%s'", flowName)
+
+	isOnlineProducer := authToken.UserID != policy.ProducerUid || flowName != models.RemittanceMgaFlow
+	if authToken.IsNetworkNode && isOnlineProducer {
+		err = errors.New("cannot access to transaction")
+		return "", nil, err
+	}
```
diogo-wopta: Just to remind you that if you change the logic in the other handler to update it here too.

diogo-wopta (approved):

PR #300, opened 2024-11-25
Dev environment seed script

fpittoni:


diogo-wopta (commented):
I'll make it short, otherwise I'll end up not giving you feedback.

Do not create the users (customers) for each entry in the email-list. Those entities must be attached to the policies. Create them, with fake data, on the run when creating the policies.

For the network nodes we must also attach the warrant, and the internal data by type (agent, agency, etc) and other data, but we can check it later.

PR #305, opened 2024-12-04
Tg qbe backend main branch

fpittoni:


models/policy.go
```diff
@@ -126,6 +130,13 @@ type Policy struct {
 	AgencyUid     string `json:"agencyUid,omitempty" firestore:"agencyUid,omitempty" bigquery:"agencyUid"` // DEPRECATED
 }
 
+type DeclaredClaim struct {
+	GuaranteeSlug string          `json:"guaranteeSlug,omitempty" firestore:"guaranteeSlug,omitempty"`
+	Year          int16           `json:"year,omitempty" firestore:"year,omitempty"`
+	Total         int16           `json:"total,omitempty" firestore:"total,omitempty"`
+	History       []DeclaredClaim `json:"history,omitempty" firestore:"history,omitempty"`
+}
+
```
diogo-wopta: Why the change in the structure? Ok for using int16 for Year, but originally we had two separate structures as follows:
```golang
type DeclaredClaims struct {
    GuaranteeSlug string
    History []DeclaredClaim
}

type DeclaredClaim struct {
    Year     string
    Quantity int
    Value    int
}
```
Anyways, I'd avoid an injection of `self` i.e: DeclaredClaim contains a field of DeclaredClaim because this nesting is a problem known with Dataflow.

The idea is: a said guarantee might have an history of claims, that are described by the year, number of events and amount

Saying that and considering what has been implemented elsewhere, I'd type the value as a float64

Regarding bigquery, for now I'd leave it blank, i.e: `bigquery:"-"` to avoid having to modify the table on the fly. When we agree to the whole structure we may add those fields there too, and write a task to remember when going into production.

diogo-wopta (changes requested):

diogo-wopta (approved):

PR #317, opened 2025-01-24
TG-983 | Upload non digital policy contract

Yousef-Hammar:


broker/upload-policy-contract.go
```diff
@@ -0,0 +1,128 @@
...
+		return "", nil, err
+	}
+	file, _, err := r.FormFile("bytes")
+	if err != nil {
+		err = fmt.Errorf("error getting file from request: %v", err)
+		return "", nil, err
+	}
+	defer file.Close()
+
+	log.Printf("policyUid: %s, mimeType: %s", policyUid, mimeType)
```
diogo-wopta: remove log. PolicyUid should already be logged in the request path and the mimeType should always be application/pdf at this point

broker/upload-policy-contract.go
```diff
@@ -0,0 +1,128 @@
...
+
+	rawDoc, err := io.ReadAll(file)
+	if err != nil {
+		err = fmt.Errorf("error reading document from request: %v", err)
+		return "", nil, err
+	}
+
+	filename := fmt.Sprintf("%s/%s/"+models.ContractDocumentFormat, "temp", policyUid,
```
diogo-wopta: can we refactor this for legibility? Is not immediate that two params are needed  for the full path and another two for the actual filename

broker/upload-policy-contract.go
```diff
@@ -0,0 +1,128 @@
...
+		node := network.GetNetworkNodeByUid(policy.ProducerUid)
+		if node != nil {
+			flow = node.GetWarrant().GetFlowName(policy.Name)
+		}
+	}
+
+	if flow == models.RemittanceMgaFlow {
+		newStatus = models.PolicyStatusSign
+		newStatusHistory = newStatusHistory[:len(newStatusHistory)-1]
+	}
```
diogo-wopta: do we need some extra steps in this case? The signed state is the last one in the remittance flow. Do we need to create the user and update the network node or those steps are already done at this point?

Yousef-Hammar: No extra steps are needed, user is created at emit time, but we need to upload contract to the permanent folder in case of remittance flow

diogo-wopta (approved):

PR #319, opened 2025-01-28
Export GDrive qbe spreadsheet and save to bucket

fpittoni:


quote/spreadsheet.go
```diff
@@ -64,8 +72,51 @@ func (qs *QuoteSpreadsheet) Spreadsheets() []Cell {
...
+
+	ssRes, _ := sheetClient.Spreadsheets.Get(qs.Id).Context(ctx).Do()
+	for _, s := range ssRes.Sheets {
+		if s.Properties.Title == qs.SheetName {
+			originalSheetId = s.Properties.SheetId
+		}
+		if s.Properties.Title == "Export" {
+			exportSheetId = s.Properties.SheetId
+		}
+	}
```
diogo-wopta: do we need this still? Are these values dynamic? If not, we could inject them in the `*QuoteSpreadsheet`. In the commercial-combined file it could then be read from the environment.

quote/spreadsheet.go
```diff
@@ -39,7 +45,9 @@ func SpreadsheetsFx(w http.ResponseWriter, r *http.Request) (string, interface{}
 
 func (qs *QuoteSpreadsheet) Spreadsheets() []Cell {
 	var (
-		path []byte
+		path               []byte
+		destinationSheetId = "1tMi7NYFZu7AnV4WkVrD0yzy1Dt3d-wVs0iZwlOcxLrg"
```
diogo-wopta: make this destinationSheetId an injected param from `*QuoteSpreadsheet`

quote/spreadsheet.go
```diff
@@ -64,8 +72,51 @@ func (qs *QuoteSpreadsheet) Spreadsheets() []Cell {
...
+		if s.Properties.Title != "Sheet1" {
+			ds := sheets.DeleteSheetRequest{SheetId: s.Properties.SheetId}
+			sr := sheets.Request{DeleteSheet: &ds}
+			sheetsToClear = append(sheetsToClear, &sr)
+		}
+	}
+	_, err := sheetClient.Spreadsheets.BatchUpdate(destinationSheetId, &sheets.BatchUpdateSpreadsheetRequest{Requests: sheetsToClear}).Context(ctx).Do()
+	if err != nil {
+		log.Printf("unable to delete sheets from spreadsheet: %v", err)
+	}
```
diogo-wopta: refactor it into a function so that it may be used like:
```golang
cleanUpSpreadsheet(client, id, func(s *Sheet) bool { return s.Properties.Title != "Sheet1"})
```

quote/spreadsheet.go
```diff
@@ -64,8 +72,51 @@ func (qs *QuoteSpreadsheet) Spreadsheets() []Cell {
...
+		DestinationSpreadsheetId: destinationSheetId,
+	}).Context(ctx).Do()
+
+	// load from drive and save to bucket
+	doc, err := loadFromDrive(path, ctx, destinationSheetId)
+	if err != nil {
+		log.Printf("unable to load from GDrive: %v", err)
+		return res
+	}
+	err = saveToBucket(bucketSavePath+"proposta_"+time.Now().Format("2006-1-2_15:4:5")+".xls", doc)
```
diogo-wopta: was this file naming agreed?
I'd suggest something like `{path}/quote_{productName}_{policyUid}_{timestamp}`

quote/spreadsheet.go
```diff
@@ -208,3 +259,46 @@ func CopySpreadsheet(path []byte, ctx context.Context, id string) (string, error
...
+		return fmt.Errorf("error uploading to bucket: %v", err)
+	}
+
+	// Set Content-Type, otherwise it will be saved as application/zip
+	err = lib.SetGoogleStorageObjectContentType(path, "application/vnd.ms-excel")
+	if err != nil {
+		return fmt.Errorf("error setting content type for obj %s: %v", path, err)
+	}
+	return nil
+}
```
diogo-wopta: since we need to set the content-type, isn't there a way to do it when uploading it? So that we may avoid the round trip (upload, retrieve, update)?

diogo-wopta (changes requested):

quote/spreadsheet.go
fpittoni: The actual save path is function-data/test/download.
Do you want to change it? Like maybe saving the spreadsheet together with the policy proposal?
If so, which should be the correct path?

lib/storage.go
```diff
@@ -32,6 +32,23 @@ func PutToGoogleStorage(bucketname string, path string, file []byte) (string, er
 	return "gs://" + bucketname + "/" + path, e
 
 }
+
+func PutToGoogleStorageWithSpecificContentType(bucketName string, path string, file []byte, contentType string) (str string, err error) {
+	log.Println("PutToGoogleStorage")
```
diogo-wopta: remove log

quote/combined-qbe.go
```diff
@@ -37,13 +37,16 @@ func CombinedQbeFx(w http.ResponseWriter, r *http.Request) (string, interface{},
...
-		InitCells:   resetCells(),
-		SheetName:   "Input dati Polizza",
+		Id:                 "1tn0Jqce-r_JKdecExFOFVEJdGUaPYdGo31A9FOgvt-Y",
+		DestinationSheetId: "1tMi7NYFZu7AnV4WkVrD0yzy1Dt3d-wVs0iZwlOcxLrg",
+		InputCells:         inputCells,
+		OutputCells:        setOutputCell(),
+		InitCells:          resetCells(),
+		SheetName:          "Input dati Polizza",
+		ExportedSheetName:  "Export",
+		ExportedSheetTitle: "Exported",
```
diogo-wopta: Could we use the same param? Keep one and discard the other if possible.

quote/spreadsheet.go
```diff
@@ -16,14 +18,21 @@ import (
...
+	OutputCells        []Cell
+	InitCells          []Cell
 }
 
-func SpreadsheetsFx(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
+var (
+	exportSheetId int64
+)
+
+/*func SpreadsheetsFx(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
```
diogo-wopta: remove the comment or delete the function if unused

quote/combined-qbe.go
```diff
@@ -37,13 +37,16 @@ func CombinedQbeFx(w http.ResponseWriter, r *http.Request) (string, interface{},
...
 	}
-	outCells := qs.Spreadsheets()
+	outCells := qs.Spreadsheets(policy)
```
diogo-wopta: instead of passing down the policy as a param, what if we passed only what was needed? Ex.:
```golang
type QuoteSpreadsheet struct {
//...
exportFilePrefix string
}
qs := QuoteSpreadsheet{
//...
exportFilePrefix: fmt.Sprintf("quote_%s_%s", policy.Name, policy.Uid)
}
//...
saveToBucket(fmt.Sprintf("%s%s_%s.xls", bucketPath, qs.exportFilePrefix, time.Now().Format("...")))
```

quote/spreadsheet.go
```diff
@@ -16,14 +18,21 @@ import (
...
+	DestinationSheetId string
+	InputCells         []Cell
+	OutputCells        []Cell
+	InitCells          []Cell
 }
 
-func SpreadsheetsFx(w http.ResponseWriter, r *http.Request) (string, interface{}, error) {
+var (
+	exportSheetId int64
+)
```
diogo-wopta: it seems this variable doesn't need to be global anymore

diogo-wopta (approved):

PR #324, opened 2025-02-05
TG-847 | Commercial combined sellable rules

fpittoni:


sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
 	"net/http"
+	"slices"
+
+	"github.com/wopta/goworkspace/lib"
+	"github.com/wopta/goworkspace/models"
 )
 
+type Out struct {
+	Msg string
+}
```
diogo-wopta: this shouldn't be public. At least not for now.
Make it private or declare it in the function where it is used.

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+	log.Println("[CommercialCombinedFx] handler start ----------- ")
+
+	body := lib.ErrorByte(io.ReadAll(r.Body))
+	defer r.Body.Close()
+	log.Printf("[CommercialCombinedFx] body: %s", string(body))
+
+	err = json.Unmarshal(body, &policy)
+	if err != nil {
+		return "", nil, err
+	}
```
diogo-wopta: Use the pattern from other handlers:
```golang
         log.SetPrefix("[CommercialCombinedFx] ")
	defer func() {
		r.Body.Close()
		if err != nil {
			log.Printf("error: %s", err.Error())
		}
		log.Println("Handler end ----------------------------------------------")
		log.SetPrefix("")
	}()

	log.Println("Handler start -----------------------------------------------")
```

Prefer the simpler syntax for reading the body:
```golang
if err = json.NewDecoder(r.Body).Decode(policy); err != nil {
		log.Printf("error decoding request body: %s", err)
		return "", nil, err
	}
```

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+	return "", nil, fmt.Errorf("policy not sellable by: %v", out.Msg)
+}
+
+func getCommercialCombinedInputData(policy *models.Policy) ([]byte, error) {
+	var numEmp = 0
+	var numBuild = 0
+	var missingMandatoryWarrant = false
+	var mandatoryThirdParty = false
+	var buildingAndRental = false
+	var mandatoryWarrantList = []string{"building", "rental-risk", "machinery", "stock"} // "third-party-liability-work-providers"
```
diogo-wopta: remove comment

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+				}
+			}
+		}
+		if v.Type == models.AssetTypeBuilding {
+			numBuild++
+			var building, rentalRisk = false, false
+			for _, g := range v.Guarantees {
+				if !slices.Contains(mandatoryWarrantList, g.Slug) {
+					missingMandatoryWarrant = true
+				}
```
diogo-wopta: possible logic error:
what if the first guarantee for the building is not one of the required (would set `missingMandatoryWarrant = true`), but the next one is (there is no setting to false again)

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+	err = json.Unmarshal(body, &policy)
+	if err != nil {
+		return "", nil, err
+	}
+
+	policy.Normalize()
+
+	in, err := getCommercialCombinedInputData(policy)
+	if err != nil {
+		log.Printf("[Commercial-Combined] error getting input data: %s", err.Error())
```
diogo-wopta: after setting the log prefix there is no need to repeat the [HandlerName] part. Keep only the necessary message.

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+
+	rulesFile := lib.GetRulesFileV2(policy.Name, policy.ProductVersion, rulesFilename)
+
+	fx := new(models.Fx)
+
+	var out = new(Out)
+	_, ruleOutput := lib.RulesFromJsonV2(fx, rulesFile, out, in, nil)
+	out = ruleOutput.(*Out)
+
+	log.Println("[CommercialCombinedFx] handler end ----------------")
```
diogo-wopta: see the first comment that sets this into the defer

sellable/commercial-combined.go
```diff
@@ -1,10 +1,108 @@
...
+
+	if out.Msg == "" {
+		return http.StatusText(http.StatusOK), nil, nil
```
diogo-wopta: return an empty json object, otherwise it won't work with the current implementation on the FE/Krakend, so:
```golang
return "{}", nil, nil
```

diogo-wopta (changes requested):
If possible, please rename your branch, using english and keeping to regular characters (avoid "à" for instance).
Github shows a warning about it that wouldn't be nice to discover that it could break things on the run.

diogo-wopta (commented):
Seems good to me. You can proceed with renaming the branch so we merge it. Open up another PR if needed.

PR #325, opened 2025-02-06
TG-847 | QBE sellable rules

diogo-wopta:


sellable/commercial-combined.go
```diff
@@ -1,10 +1,109 @@
...
\ No newline at end of file
+	var (
+		policy *models.Policy
+		err    error
+	)
+	log.SetPrefix("[CommercialCombinedFx]")
+	log.Println("handler start ----------- ")
+
+	defer func() {
+		err = r.Body.Close()
```
diogo-wopta: @fpittoni do not overwrite the err with the one from body close, otherwise we'll empty out the one thrown by the handler. Since we are in a Cloud Function, you can ignore its error.

PR #327, opened 2025-02-07
New data field added to Transaction; modified BigQueryParse() accordingly

fpittoni:


models/transaction.go
```diff
@@ -104,6 +105,13 @@ func (t *Transaction) Normalize() {
 }
 
 func (t *Transaction) BigQueryParse() {
+	data, err := json.Marshal(t)
+	if err != nil {
+		log.Println("ERROR Transaction "+t.Uid+" Marshal: ", err)
```
diogo-wopta: use `log.Printf()`

models/transaction.go
```diff
@@ -114,16 +122,11 @@ func (t *Transaction) BigQueryParse() {
 
 func (t *Transaction) BigQuerySave(origin string) {
 	fireTransactions := lib.GetDatasetByEnv(origin, TransactionsCollection)
```
diogo-wopta:
```golang
fireTransactions := lib.GetDatasetByEnv(origin, TransactionsCollection)
```
since we're here clean this up as it's not needed

models/transaction.go
```diff
@@ -114,16 +122,11 @@ func (t *Transaction) BigQueryParse() {
...
-		log.Println("ERROR Transaction "+t.Uid+" Marshal: ", err)
-		return
-	}
-	log.Println("Transaction: "+t.Uid, string(transactionJson))
+
 	t.BigQueryParse()
 	log.Println("Transaction save BigQuery: " + t.Uid)
 
-	err = lib.InsertRowsBigQuery(WoptaDataset, fireTransactions, t)
+	err := lib.InsertRowsBigQuery(WoptaDataset, fireTransactions, t)
```
diogo-wopta: replace `fireTransactions` with `lib.TransactionsCollection` (or something like it)

diogo-wopta (changes requested):

diogo-wopta:
In this PR we must include a script for modifying the present data so that all transactions on BigQuery will have the updates Data value.

The idea is to get all transactions from firestore and save them to bigquery to force having the Data field compiled.

_script/copy_transactions_to_bigquery.go
```diff
@@ -26,3 +32,49 @@ func CopyTransactionsToBigQuery() {
...
+		}
+
+		var policy models.Policy
+		err = doc.DataTo(&policy)
+		if err != nil {
+			return fmt.Errorf("unable to populate policy: %w", err)
+		}
+
+		policyUids = append(policyUids, policy.Uid)
+	}
```
diogo-wopta: I'll have to check with Luca/Linda if we need to update all transactions, or only those of non deleted policies as you have done.

_script/copy_transactions_to_bigquery.go
```diff
@@ -26,3 +32,49 @@ func CopyTransactionsToBigQuery() {
...
+
+	for _, uid := range policyUids {
+		transactions := transaction.GetPolicyTransactions("", uid)
+
+		for index, tr := range transactions {
+			t := deepcopy.Copy(tr).(models.Transaction)
+			t.BigQuerySave("")
+			transactions[index].BigQuerySave("")
+		}
+	}
```
diogo-wopta: This way we are making a DB request for each transaction. If you pass a list of the same type to the function `lib.InsertRowsBigQuery` you can do it in a single batch. Ex.:
```golang
lib.InsertRowsBigQuery({dataset}, {collection}, []models.Transaction{............})
```

The batch size could be by policy, or a number n of policies.

diogo-wopta (changes requested):
In the script, add at least a log to signal that the script has ended. We usually do not return errors from those script but exit the program.

diogo-wopta:
I still have to check the last commits, but talking with Luca we came up with the idea of updating only the PAID transactions. This way we wouldn't need to suspend the operations.

_script/copy_transactions_to_bigquery.go
```diff
@@ -26,3 +32,75 @@ func CopyTransactionsToBigQuery() {
...
+	log.Println("Finished copying all transactions to BigQuery")
+}
+
+func divideSliceIntoBatches[T any](slice []T, size int) ([][]T, error) {
+	if size < 1 {
+		return nil, errors.New("size must be greater than zero")
+	}
+	if len(slice) < size {
+		return nil, errors.New("size must be smaller than len(slice)")
+	}
```
diogo-wopta: this shouldn't be an error. If the size is 100 and the length of the slice is 80, we should get back a list of batches with one single batch with all 80 elements.

`[][]T{[]T{80...}}`

_script/copy_transactions_to_bigquery.go
diogo-wopta: just for book keeping, we'll work with all paid transactions

diogo-wopta (approved):

PR #330, opened 2025-02-14
No tg qbe sellable fix

fpittoni:


policy/update-policy.go
```diff
@@ -38,10 +38,12 @@ func UpdatePolicy(policy *models.Policy) (map[string]interface{}, error) {
...
-		err := checkQbe(policy, input)
-		if err != nil {
-			return nil, err
-		}
+		// NOTE: remove qbe check for now.
+
+		//err := checkQbe(policy, input)
+		//if err != nil {
+		//	return nil, err
+		//}
```
diogo-wopta: we cannot do it like this. If you comment out the whole function it won't add to the input map the needed values. We have to turn-off only the validations.

diogo-wopta (changes requested):

diogo-wopta (approved):

PR #333, opened 2025-02-20
TG-1029 | GetUser query-builder

fpittoni:


user/get-users.go
```diff
@@ -161,3 +163,75 @@ func GetAuthUserByMail(origin, mail string) (models.User, error) {
...
+		log.SetPrefix("")
+	}()
+	log.Println("Handler start -----------------------------------------------")
+
+	paramsMap := extractQueryParams(r)
+	if len(paramsMap) == 0 {
+		return "", nil, errors.New("no query params")
+	}
+
+	queryBuilder := qb.NewQueryBuilder("user")
```
diogo-wopta: we don't need to vary the builder, it will be always the same so we can remove the `"user"` param for now.

user/query-builder/pkg/factory.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/wopta/goworkspace/user/query-builder/internal/user"
+
+func NewQueryBuilder(tmp string) QueryBuilder {
+	switch tmp {
+	case "user":
+		return user.NewQueryBuilder(nil)
+	}
+	return nil
+}
```
diogo-wopta: No need for different queryBuilders for now, you can always return:
```golang
return user.NewQueryBuilder()
```

user/query-builder/internal/user/query-builder.go
```diff
@@ -0,0 +1,33 @@
...
+type UsersQueryBuilder struct {
+	base.UsersQueryBuilder
+}
+
+func NewQueryBuilder(randomGenerator func() string) *UsersQueryBuilder {
+	return &UsersQueryBuilder{
+		base.NewQueryBuilder(lib.UsersViewCollection, "u", randomGenerator,
+			paramsHierarchy, paramsWhereClause),
+	}
+}
```
diogo-wopta: The policy implementation injected the randomGenerator only for testing purposes. Ideally, we should not inject it and always use the same regardless the case.

user/query-builder/internal/user/query-builder.go
```diff
@@ -0,0 +1,33 @@
...
+	//	deleteClause = "(**tableAlias**.isDeleted = false OR **tableAlias**." +
+	//		"isDeleted IS NULL)"
+	//	emitClause = "(**tableAlias**.companyEmit = true)"
+	//)
+	//qb.WhereClauses = []string{emitClause, deleteClause}
+	//if val, ok := params["status"]; ok {
+	//	if strings.Contains(val, "deleted") {
+	//		qb.WhereClauses = qb.WhereClauses[:1]
+	//	}
+	//}
```
diogo-wopta: Remove comments. If we ever change the requirements we could add a "non deleted" where clause.

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	qb.limit = uint64(limit)
+
+	return nil
+}
+
+func (qb *UsersQueryBuilder) parseQuery() string {
+	const queryPrefix = "SELECT **tableAlias**.uid, **tableAlias**.name " +
+		"FROM `wopta.**tableName**` **tableAlias** " +
+		"WHERE "
```
diogo-wopta: It seems to me that you have to add all the fields from the `type userInfo` here so we can actually extract all that data.
```golang
const queryPrefix = "SELECT **tableAlias**.uid, **tableAlias**.name,  **tableAlias**.surname, **tableAlias**.mail, **tableAlias**.role, **tableAlias**.fiscalCode FROM ...."
```

Ideally this kind of data should be declared only once. If it comes to you any idea of improving it we can discuss.

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+		rawQuery bytes.Buffer
+	)
+
+	rawQuery.WriteString(queryPrefix)
+	if len(qb.WhereClauses) == 1 {
+		rawQuery.WriteString(qb.WhereClauses[0])
+	} else if len(qb.WhereClauses) > 1 {
+		rawQuery.WriteString(strings.Join(qb.WhereClauses, " AND "))
+	}
+	rawQuery.WriteString(fmt.Sprintf(" ORDER BY **tableAlias**.updateDate DESC"))
```
diogo-wopta: the field in the User model is named `updatedDate`

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+			if err != nil {
+				return nil, nil, err
+			}
+			whereClauses = append(whereClauses, whereClause)
+		} else if paramKey == "DO_NOT_EXIST" {
+			whereClause := qb.processProducerUidParam(paramValue, queryParams)
+			if whereClause != "" {
+				whereClauses = append(whereClauses, whereClause)
+			}
+		} else {
```
diogo-wopta: We have no use for translatedKeys and it seems the `DO_NOT_EXIST` key is not set anywhere. Thus, let's remove these blocks and keep only what is declared in the `else` part.

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (qb *UsersQueryBuilder) processProducerUidParam(paramValue string, queryParams map[string]interface{}) string {
+	tmp := make([]string, 0)
+	for _, uid := range strings.Split(paramValue, ",") {
+		randomIdentifier := qb.randomGenerator()
+		queryParams[randomIdentifier] = lib.TrimSpace(uid)
+		tmp = append(tmp, fmt.Sprintf("@%s", randomIdentifier))
+	}
+	return fmt.Sprintf(qb.paramsWhereClause["mail"], strings.Join(tmp, ", "))
+}
```
diogo-wopta: Unused, remove it.

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+		if val, ok := qb.paramsWhereClause[lib.TrimSpace(status)]; ok && val != "" {
+			whereClauses = append(whereClauses, val)
+		}
+	}
+	if len(whereClauses) == 0 {
+		return "", errors.New("error processing params")
+	}
+
+	return "(" + strings.Join(whereClauses, " OR ") + ")", nil
+}
```
diogo-wopta: Unused, remove it.

user/query-builder/internal/base/query-builder.go
```diff
@@ -0,0 +1,213 @@
...
+			rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+			s := make([]rune, 12)
+			s[0] = letters[rnd.Intn(len(letters))]
+			for i := range s[1:] {
+				s[i+1] = alphanum[rnd.Intn(len(alphanum))]
+			}
+			return string(s)
+		}
+	}
```
diogo-wopta: Consider that we shouldn't inject the randomGenerator, so:
1- declare and use it here inside the constructor, or;
2- declare it outside and use it here;

user/get-users.go
```diff
@@ -161,3 +163,75 @@ func GetAuthUserByMail(origin, mail string) (models.User, error) {
...
+	if err != nil {
+		return "", nil, fmt.Errorf("error generating query: %v", err)
+	}
+
+	log.Println("paramsMap:")
+	log.Println(paramsMap)
+	log.Println("QUERY:")
+	log.Println(query)
+	log.Println("QUERY PARAMS:")
+	log.Println(queryParams)
```
diogo-wopta: For standard, use the same format as the portfolio handler:
```golang
log.Printf("query: %s\nqueryParams: %+v", query, queryParams)
```

user/get-users.go
```diff
@@ -161,3 +163,75 @@ func GetAuthUserByMail(origin, mail string) (models.User, error) {
...
+		log.Println("Handler end ---------------------------------------------")
+		log.SetPrefix("")
+	}()
+	log.Println("Handler start -----------------------------------------------")
+
+	paramsMap := extractQueryParams(r)
+	if len(paramsMap) == 0 {
+		return "", nil, errors.New("no query params")
+	}
+
```
diogo-wopta, line 198: For standard with the portfolio handler add log:
```golang
log.Printf("input params: %v", paramsMap)
```

user/get-users.go
```diff
@@ -12,6 +13,7 @@ import (
 
 	lib "github.com/wopta/goworkspace/lib"
 	models "github.com/wopta/goworkspace/models"
```
diogo-wopta: Not related to your chages but fix the import: no need to rename the autogenerated package name
```golang
"github.com/wopta/goworkspace/lib"
"github.com/wopta/goworkspace/models"
```

diogo-wopta (changes requested):

diogo-wopta (approved):
It seems good to me. I'll approve it but let's merge it only after the planned release so we can test it with the correct view.

PR #334, opened 2025-02-25
Added a new function to modify Beneficiaries and BeneficiaryReference

fpittoni:


models/constants.go
```diff
@@ -218,6 +218,7 @@ const (
 	FlowFileFormat                = "flows/%s.json"
 	ContractDocumentFormat        = "%s_Contratto_%s.pdf"
 	ProposalDocumentFormat        = "%s_Proposta_%d.pdf"
+	AddendumDocumentFormat        = "%s_Appendice_%d.pdf"
```
diogo-wopta: Review format: a policy could have multiples addendums

mga/modify-policy.go
```diff
@@ -6,9 +6,11 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"reflect"
```
diogo-wopta, line 9: Can we compare data the same way? Why use reflect.DeepEqual for some and field by field comparison for others?

mga/modify-policy.go
```diff
@@ -67,12 +69,38 @@ func ModifyPolicyFx(w http.ResponseWriter, r *http.Request) (string, interface{}
...
+	err = writePolicyToDb(modifiedPolicy)
+	if err != nil {
+		return "{}", nil, err
+	}
+	log.Printf("policy %s successfully saved", modifiedPolicy.Uid)
+	err = writeUserToDB(modifiedUser)
+	if err != nil {
+		return "{}", nil, err
+	}
+	log.Printf("user %s modified successfully", modifiedUser.Uid)
```
diogo-wopta, line 107: We should write to the DB only at the end, otherwise will lose the new attachment or have to save twice.

mga/modify-policy.go
```diff
@@ -81,7 +109,154 @@ func ModifyPolicyFx(w http.ResponseWriter, r *http.Request) (string, interface{}
...
-func modifyController(originalPolicy, inputPolicy models.Policy) (models.Policy, error) {
+func generateDiffPolicy(originalPolicy, inputPolicy models.Policy) (models.Policy, error) {
+	var diff models.Policy
+	diff.Uid = originalPolicy.Uid
+	diff.Name = originalPolicy.Name
+	diff.NameDesc = originalPolicy.NameDesc
+	diff.ProposalNumber = originalPolicy.ProposalNumber
+	diff.StartDate = originalPolicy.StartDate
+	diff.EndDate = originalPolicy.EndDate
+	diff.Company = originalPolicy.Company
```
diogo-wopta: It seems to me that this block should be placed at the end, and only if there is an actual diff. What would happen if someone invoked the endpoint with the same data, without altering it? It seems we would generate an addendum with no data for the tables..

If needed you could track if there is an actual diff in a boolean and return it. From outside you could call the document api only if there is need for it.

mga/modify-policy.go
```diff
@@ -81,7 +109,154 @@ func ModifyPolicyFx(w http.ResponseWriter, r *http.Request) (string, interface{}
...
+	}
+	if inputPolicy.Assets[0].Guarantees[0].Beneficiaries != nil {
+		if originalPolicy.Assets[0].Guarantees[0].Beneficiaries != nil {
+			gar.Beneficiaries = diffForBeneficiaries(*(originalPolicy.Assets[0].Guarantees[0]).Beneficiaries, *(inputPolicy.Assets[0].Guarantees[0]).Beneficiaries)
+		} else {
+			gar.Beneficiaries = inputPolicy.Assets[0].Guarantees[0].Beneficiaries
+		}
+	} else {
+		gar.Beneficiaries = nil
+	}
```
diogo-wopta: for now we can get away accessing the Assets slice on index 0 but that is not extendable. 
Whenever there is and index 0 access try making a loop and checking for the data

document/addendum.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/wopta/goworkspace/models"
+)
+
+type AddendumResponse struct {
+	LinkGcs  string `json:"linkGcs"`
+	Filename string `json:"fileName"`
+}
+
+func AddendumObj(origin string, data models.Policy, networkNode *models.NetworkNode, product *models.Product) <-chan AddendumResponse {
+	r := make(chan AddendumResponse)
```
diogo-wopta: I know the other implementations are like this, but do we need to make this in a  go routine?

document/addendum.go
```diff
@@ -0,0 +1,96 @@
...
+
+	go func() {
+		var (
+			err      error
+			filename string
+			out      []byte
+		)
+
+		switch data.Name {
+		case models.PmiProduct:
```
diogo-wopta: remove the unhandled cases and place a default that at least makes a log for unimplemented method, perhaps trowing an error that could be checked from outside like:
```golang
errors.is(err, errAddendumUnimplemented)
```

document/addendum.go
```diff
@@ -0,0 +1,96 @@
...
+			Filename: filename,
+		}
+	}()
+
+	log.Println("[AddendumObj] function end -------------------------------..")
+
+	return r
+}
+
+/*func GetGenerator(origin string, data models.Policy, networkNode *models.NetworkNode, product *models.Product) {
```
diogo-wopta: delete commented code

document/addendum.go
```diff
@@ -0,0 +1,96 @@
...
+			out      []byte
+		)
+
+		switch data.Name {
+		case models.PmiProduct:
+
+		case models.LifeProduct:
+			prod := models.Product{}
+			generator := addedndum.NewLifeAddendumGenerator(engine.NewFpdf(), &data, networkNode, prod)
+			out, err = generator.Contract()
```
diogo-wopta: rename method. This is not a contract. It could be alway something like `Generate()`

document/internal/dto/beneficiaries-dto.go
diogo-wopta, line 1: general comments:
- rename file;
- rename types: they don't seem life-product-only
- avoid Public access whenever possible. Some of these types are used only in this internal context so should be private

document/internal/dto/beneficiaries-dto.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+}
+
+func (b *BeneficiariesDTO) FromPolicy(policy models.Policy, product models.Product) {
+	b.Contract.fromPolicy(policy)
+	b.Contractor.fromPolicy(policy.Contractor)
+	b.Insured.fromPolicy(policy.Assets[0].Person)
+	//b.Beneficiaries.fromPolicy(policy.Assets[0].Guarantees)
+	b.Beneficiaries.fromPolicy(policy.Assets[0].Guarantees[0].Beneficiaries, policy.Assets[0].Guarantees[0].BeneficiaryOptions)
+	b.BeneficiaryReference.fromPolicy(policy.Assets[0].Guarantees[0].BeneficiaryReference)
```
diogo-wopta: same as another comment further below:
avoid accessing by index 0. Loop the slices and select what is needed. for now we could exit on the first match, in the future we might need to keep track of all assets changes

document/internal/dto/beneficiaries-dto.go
```diff
@@ -0,0 +1,319 @@
...
+		DomStreetNumber: constants.EmptyField,
+		DomCity:         constants.EmptyField,
+		DomProvince:     constants.EmptyField,
+		Mail:            constants.EmptyField,
+		Phone:           constants.EmptyField,
+	}
+}
+
+func newLifeBeneficiariesDTO() *Beneficiaries {
+	lb := make([]lifeBeneficiaryDTO, 0)
```
diogo-wopta: if you make here is of the same type as the return type there won't be a need for creating an extra variable on line 173
`lb := make(Beneficiaries, 0)`

document/internal/dto/beneficiaries-dto.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	if contr.Mail != "" {
+		lc.Mail = contr.Mail
+	}
+	if contr.Phone != "" {
+		lc.Phone = contr.Phone
+	}
+}
+
+func (li *lifeInsuredDTO) fromPolicy(ins *models.User) {
```
diogo-wopta: Attention, you are accessing a pointer without checking it for nil, it could panic

document/internal/dto/beneficiaries-dto.go
```diff
@@ -0,0 +1,319 @@
...
+			BirthDate:    (*b)[i].BirthDate,
+			Contactable:  v.IsContactable,
+			Relation:     "\n" + buf.String(),
+		}
+		(*b)[i] = ben
+	}
+}
+
+func (br *BeneficiaryReferenceDTO) fromPolicy(benRef *models.User) {
+	if benRef.FiscalCode != "" {
```
diogo-wopta: Attention, you are accessing a pointer without checking it for nil, it could panic (it happened to me locally)

document/internal/dto/enterprise-dto.go
```diff
@@ -39,6 +39,9 @@ func (e *enterpriseDTO) fromPolicy(enterprise models.Enterprise, guarantees []mo
 	}
 
 	for _, guarantee := range guarantees {
-		e.Guarantees[guarantee.Slug].fromPolicy(guarantee)
+		if val, ok := e.Guarantees[guarantee.Slug]; ok {
+			val.fromPolicy(guarantee)
+		}
+
```
diogo-wopta, line 45: I tend to agree with the change, but why it was needed?

document/pkg/addedndum/base.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	tabDimension = 15
+)
+
+type baseGenerator struct {
+	engine      *engine.Fpdf
+	now         time.Time
+	signatureID uint32
```
diogo-wopta: no integration with Namirial is needed so signatureID won't be used

document/pkg/addedndum/base.go
```diff
@@ -0,0 +1,113 @@
...
+	networkNode *models.NetworkNode
+	policy      *models.Policy
+}
+
+func (bg *baseGenerator) Save(rawDoc []byte) (string, error) {
+	filename := strings.ReplaceAll(fmt.Sprintf("%s/%s/"+models.AddendumDocumentFormat, "temp", bg.policy.Uid,
```
diogo-wopta: do not save the file on temp, this is an actual policy
replace your "%s/%s/" with
```golang
userPathFormat = "assets/users/%s/"
fmt.Sprintf(userPathFormat, policy.Contractor.Uid)
```

check the comment on the last file, we need to improve the naming

document/pkg/addedndum/base.go
```diff
@@ -0,0 +1,113 @@
...
+func (bg *baseGenerator) woptaFooter() {
+	const (
+		rowHeight   = 3
+		columnWidth = 50
+	)
+
+	bg.engine.SetFooter(func() {
+		bg.engine.SetY(-30)
+
+		//currentY := bg.engine.GetY()
```
diogo-wopta: remove commented code

document/pkg/addedndum/life.go
```diff
@@ -0,0 +1,878 @@
...
+		dto: LifeAddendumDTO,
+	}
+}
+
+func (lag *LifeAddendumGenerator) Contract() ([]byte, error) {
+	lag.mainHeader()
+
+	lag.engine.NewPage()
+
+	lag.WoptaPerTe()
```
diogo-wopta: this call should be inside mainHeader. What if in the future we'll have a second page?

diogo-wopta (changes requested):
I haven't checked the whole addendum life file, but there are already some changes to consider

document/internal/dto/enterprise-dto.go
fpittoni: First time I tried to run the code (to gain insight about pdf generation), triggering the endpoint with the policy used by postman caused a panic. I didn't investigate that much but, IIRC, it tried to use a guarantee slug that didn't exist.

diogo-wopta:
@fpittoni merge master to check for possible conflicts and fix pipeline, it seems you have an unused import. Other then that check the naming for the package/module `addendum` there are some typos.

diogo-wopta:
Will be merged in gitlab

PR #344, opened 2025-03-24
Tg 1023 cover life

diogo-wopta:


diogo-wopta:
Will be merged in gitlab

